PR #1, opened 2022-01-20
feat: add gvisor

andrewrynhard:
Signed-off-by: Andrew Rynhard <andrew@rynhard.io>

### DIFF
--- a/gvisor/pkg.yaml
+++ b/gvisor/pkg.yaml
@@ -1,0 +1,50 @@
+name: gvisor
+variant: scratch
+shell: /toolchain/bin/bash
+dependencies:
+  - stage: base
+steps:
+  - sources:
+      # sync with commit in build
+      - url: https://github.com/google/gvisor/archive/c1512ec8067c772473a4d6bad12953848eab8552.tar.gz
+        destination: gvisor.tar.gz
+        sha256: df41a38cc4d6068e6475f2f0a29c083bf11fd682869957b1b325980d3892725b
+        sha512: f4fd4bd5fbf482fceb7c46311c5dfd808cbf40e921c5f92ce4011f9b58e477af0dc3d5da8e1175c7ff22ad42d8351d6e8b9d8cc4d8339ded7cd782d617331002
+    env:
+      GOPATH: /go
+    prepare:
+      - |
+        mkdir -p /etc/ssl/certs/
+        ln -s /toolchain/etc/ssl/certs/ca-certificates /etc/ssl/certs/ca-certificates
+
+        mkdir -p ${GOPATH}/src/github.com/google/gvisor
+
+        tar -xzf gvisor.tar.gz --strip-components=1 -C ${GOPATH}/src/github.com/google/gvisor
+    build:
+      - |
+        export PATH=${PATH}:${TOOLCHAIN}/go/bin
+        cd ${GOPATH}/src/github.com/google/gvisor
+
+        mkdir ./bin
+
+        CGO_ENABLED=0 go build -o ./bin/runsc ./runsc
+
+        CGO_ENABLED=0 go build -o ./bin/containerd-shim-runsc-v1 ./shim
+    install:
+      - |
+        mkdir -p /rootfs/usr/local/bin
+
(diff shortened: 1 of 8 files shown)


gvisor/pkg.yaml
```diff
@@ -0,0 +1,49 @@
...
+
+        cp ./bin/runsc /rootfs/rootfs/usr/local/bin/runsc
+        chmod +x /rootfs/rootfs/usr/local/bin/runsc
+
+        cp ./bin/containerd-shim-runsc-v1 /rootfs/rootfs/usr/local/bin/containerd-shim-runsc-v1
+        chmod +x /rootfs/rootfs/usr/local/bin/containerd-shim-runsc-v1
+
+        cp /pkg/manifest.yaml /rootfs/
+finalize:
+  - from: /rootfs
```
smira, line 47: we need `manifest.yaml` at `/` and extension filesystem at `/rootfs`

gvisor/pkg.yaml
```diff
@@ -0,0 +1,49 @@
...
+      - |
+        mkdir -p /rootfs/rootfs/usr/local/bin
+
+        cd ${GOPATH}/src/github.com/google/gvisor
+
+        cp ./bin/runsc /rootfs/rootfs/usr/local/bin/runsc
+        chmod +x /rootfs/rootfs/usr/local/bin/runsc
+
+        cp ./bin/containerd-shim-runsc-v1 /rootfs/rootfs/usr/local/bin/containerd-shim-runsc-v1
+        chmod +x /rootfs/rootfs/usr/local/bin/containerd-shim-runsc-v1
```
smira, line 43: we also need a config file for containerd to enable `runsc` as an optional container runtime

it should go to `/etc/cri/conf.d/gvisor.part` and it will be merged by Talos to the final CRI config (https://github.com/talos-systems/talos/pull/4825/files)

gvisor/pkg.yaml
```diff
@@ -0,0 +1,49 @@
...
+        CGO_ENABLED=0 go build -o ./bin/runsc ./runsc
+
+        CGO_ENABLED=0 go build -o ./bin/containerd-shim-runsc-v1 ./shim
+    install:
+      - |
+        mkdir -p /rootfs/rootfs/usr/local/bin
+
+        cd ${GOPATH}/src/github.com/google/gvisor
+
+        cp ./bin/runsc /rootfs/rootfs/usr/local/bin/runsc
```
andrewrynhard, line 39: @smira Notice `/rootfs/rootfs`. `manifest.yaml` goes into `/rootfs` and everything else is under `/rootfs/rootfs`.

smira: ah, ok, that is a bit confusing, and I missed that. 

should we probably rename one of these? or is it too weird?

andrewrynhard: Yeah we had this same thought and figured we can have that debate here 🙂 . I threw out the idea of something like `/rootfs/extension` or `/rootfs/extensionfs`.

smira: or could we `finalize` in two steps, `/rootfs` -> `/rootfs`, `/manifest.yaml` -> `/manifest.yaml` ?

smira: I want to avoid nested `/rootfs/rootfs`

andrewrynhard: I like that idea.

smira (approved):

gvisor/pkg.yaml
```diff
@@ -0,0 +1,52 @@
...
+
+        cd ${GOPATH}/src/github.com/google/gvisor
+
+        cp ./bin/runsc /rootfs/usr/local/bin/runsc
+        chmod +x /rootfs/usr/local/bin/runsc
+
+        cp ./bin/containerd-shim-runsc-v1 /rootfs/usr/local/bin/containerd-shim-runsc-v1
+        chmod +x /rootfs/usr/local/bin/containerd-shim-runsc-v1
+
+        cp  /rootfs/
```
smira, line 45: remove me?

PR #3, opened 2022-01-24
feat: add extension for intel-ucode

rsmitty:
This PR adds in the ability to install the intel ucode binaries as a
system extension.

Signed-off-by: Spencer Smith <spencer.smith@talos-systems.com>

### DIFF
--- a/intel-ucode/pkg.yaml
+++ b/intel-ucode/pkg.yaml
@@ -1,0 +1,23 @@
+name: intel-ucode
+variant: scratch
+shell: /toolchain/bin/bash
+dependencies:
+  - stage: base
+steps:
+  - sources:
+      - url: https://github.com/intel/Intel-Linux-Processor-Microcode-Data-Files/archive/refs/tags/microcode-20210608.tar.gz
+        destination: intel-ucode.tar.gz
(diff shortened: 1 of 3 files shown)


andrewrynhard (approved):

frezbo (approved):

intel-ucode/pkg.yaml
```diff
@@ -0,0 +1,23 @@
...
+        destination: intel-ucode.tar.gz
+        sha256: fd85b6b769efd029dec6a2c07106fd18fb4dcb548b7bc4cde09295a8344ef6d7
+        sha512: 61acd2e76aa019fa0002fbf56c503791080a937ff93d81e020f8f0cc089dc08928b4c7e9884f713b886e2f9d4a8409fea59e39f628ef534a588515e1c3fc861d
+    prepare:
+      - |
+        tar -xzf intel-ucode.tar.gz --strip-components=1
+    install:
+      - |
+        mkdir -p /rootfs/lib/firmware
+        cp -R intel-ucode /rootfs/lib/firmware
```
smira, line 18: I think we should add `-p` to preserve the timestamps

smira (approved):
:+1: should we do same for AMD, looks like they have something in linux-firmware: https://git.kernel.org/pub/scm/linux/kernel/git/firmware/linux-firmware.git/tree/amd-ucode

PR #5, opened 2022-01-25
feat: update gvisor extension after testing

smira:
This extension was tested with Talos, there's one issue with cgroups
which required disabling cgroup support in `runsc`.

Fixed up manifest and added documentation on using the extension.

Signed-off-by: Andrey Smirnov <andrey.smirnov@talos-systems.com>

### DIFF
--- a/gvisor/README.md
+++ b/gvisor/README.md
@@ -1,0 +1,53 @@
+# gVisor extension
+
+## Usage
+
+Enable the extension in the machine configuration before installing Talos:
+
+```yaml
+machine:
+  install:
+    extensions:
+      - image: ghcr.io/talos-systems/gvisor:<VERSION>
+```
+
+gVisor requires unprivileged user namespace creation, so Talos default setting
+should be overridden:
+
+```yaml
+machine:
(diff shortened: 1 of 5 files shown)


gvisor/README.md
```diff
@@ -0,0 +1,53 @@
...
+## Testing
+
+Apply the following manifest to run nginx pod via gVisor:
+
+```yaml
+apiVersion: node.k8s.io/v1
+kind: RuntimeClass
+metadata:
+  name: gvisor
+handler: runsc
```
andrewrynhard, line 34: First time I have seen a Kubernetes resource have something other than metadata or spec at this level in the YAML. Just pointing this out in case it is a typo.

andrewrynhard (approved):

frezbo (approved):

gvisor/README.md
smira: it works :man_shrugging: that's the only thing I can say in my defense :-D

Unix4ever (approved):

PR #14, opened 2022-02-10
feat: add mdadm support

sergelogvinov:
Activate healthy mdraid at boot time.

Signed-off-by: Serge Logvinov <serge.logvinov@sinextra.dev>


internal/eudev/pkg.yaml
```diff
@@ -0,0 +1,46 @@
+name: eudev
```
smira, line 1: I feel we need a better story here, I don't want to duplicate every pkg we have in `pkgs/` already. I will think what we can do to provide better story here

frezbo: I guess we can depend on the published image from `pkgs`

smira: yep, exactly, I just wanted to to a bit of refactoring to make this easier

dm3ch:
Am I right understood that this PR would allow having a root partition on software raid (mirror)?

By the way, I see that there were no activity for almost a year, so what is the actual status/plans? Maybe some help from outside needed?

Gottox:
What's currently blocking? - I think I could make use of this extension ^^

sergelogvinov:
Hello, I think this extension can be used only for csi-drivers (for non system disk)
For system disk we need to add support for boot loader and add some (huge) code in Talos.

Probably you can create raid manually and `dd` image to the soft raid after. But i did not try it.
I believe that mdadm binary should be a part of the system as lvm utils (which already there).

PS. only 2 like in this pr, so is not so popular...

sergelogvinov:
new implementation https://github.com/siderolabs/extensions/pull/271

sergelogvinov:
https://github.com/siderolabs/extensions/pull/271 was merged

PR #18, opened 2022-02-18
feat: add FRR

andrewrynhard:


andrewrynhard:
I don't see how to make this a draft after the fact :D

netnoddy:
We are trying to build a Talos image with FRR as a service extension is this supported?

vbalexr:
I would like to help if hands needed!

Why are there iscsi files here? Should not the get removed?

frezbo:
Closing since talos native FRR is the way forward

PR #25, opened 2022-04-04
refactor: use `base` image, bump bldr, update versions

smira:
Update metadata for Talos 1.0.0.

Bump `bldr` to the latest version with `Merge` support, workaround issue
with `/tmp` losing sticky bit.

Use `base` image from pkgs to drop dependency on `tools`.

Signed-off-by: Andrey Smirnov <andrey.smirnov@talos-systems.com>


Unix4ever (approved):

nvidia-container-toolkit/glibc/pkg.yaml
```diff
@@ -13,6 +13,7 @@ steps:
       DEBIAN_FRONTEND: noninteractive
     prepare:
       - |
+        chmod 1777 /tmp
```
frezbo, line 16: bldr issue?

smira: yep, exactly, it resets permissions on merge (?). not sure why?

but seems harmless otherwise

PR #46, opened 2022-06-16
chore: bump nvidia driver to 515.48.07

frezbo:
Bump NVIDIA drivers to 515.48.07
Bump libnvidia-container to v1.10.0
Bump container-toolkit to v1.10.0

Ref: https://gitlab.com/nvidia/container-toolkit/libnvidia-container/-/issues/1


Unix4ever:
Looks like this hash should be be updated too https://github.com/siderolabs/extensions/blob/b5de7b49e8b39a9ade689b0ed89fda592f86ab2c/nonfree/nvidia-fabricmanager/pkg.yaml#L16 ?

frezbo:
> Looks like this hash should be be updated too

ahh I need to fix that, fabric manager seems to follow a different release versioning. It doesn't release for every nvidia driver versions, but some only.

frezbo:
Superseded by #51

PR #52, opened 2022-08-15
feat: allow modules to be loaded via extension

frezbo:
Allow modules to be loaded via extensions.


smira (approved):

frezbo:
On talos side we could have `kmod-tools` inside installer to be able to create the final module list. For now only loading a single extra module build is possible

PR #69, opened 2022-09-11
feat: add gasket driver

ammmze:
This driver enables the use of [Coral.ai](https://coral.ai/)'s PCIe TPU modules.

Signed-off-by: Branden Cash <203336+ammmze@users.noreply.github.com>

Depends-on: siderolabs/pkgs#580

### DIFF
--- a/drivers/gasket/manifest.yaml
+++ b/drivers/gasket/manifest.yaml
@@ -1,0 +1,12 @@
+version: v1alpha1
+metadata:
+  name: gasket
+  version: "$VERSION"
+  author: Sidero Labs
+  description: |
+    This system extension provides google gasket driver kernel modules built against a specific Talos version.
+    This driver is required for PCIe and M.2 Google Coral accelerators. There are 2 kernel modules ("gasket"
+    and "apex") required to enable this driver.
+  compatibility:
+    talos:
+      version: ">= v1.2.0"
--- a/drivers/gasket/vars.yaml
+++ b/drivers/gasket/vars.yaml
@@ -1,0 +1,2 @@
+# update when pkgs version is updated
+VERSION: "97aeba58-{{ .BUILD_ARG_TAG }}"
--- a/Makefile
+++ b/Makefile
@@ -35,7 +35,19 @@ COMMON_ARGS += --build-arg=TAG=$(TAG)
 empty :=
 space = $(empty) $(empty)
 
-TARGETS = amd-ucode bnx2-bnx2x gvisor hello-world-service intel-ucode iscsi-tools nut-client nvidia-container-toolkit nvidia-fabricmanager nvidia-open-gpu-kernel-modules
+TARGETS = \
+		amd-ucode \
+		bnx2-bnx2x \
+		gasket-driver \
+		gvisor \
+		hello-world-service \
+		intel-ucode \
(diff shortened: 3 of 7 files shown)


drivers/gasket/manifest.yaml
```diff
@@ -0,0 +1,12 @@
...
+  name: gasket
+  version: "$VERSION"
+  author: Sidero Labs
+  description: |
+    This system extension provides google gasket driver kernel modules built against a specific Talos version.
+    This driver is required for PCIe and M.2 Google Coral accelerators. There are 2 kernel modules ("gasket"
+    and "apex") required to enable this driver.
+  compatibility:
+    talos:
+      version: ">= v1.2.0"
```
ammmze, line 12: For kernel modules extensions, should the version be constrained to the only the precise Talos version? My understanding is that each Talos version would need to have a distinct version of the kernel modules because the signature would be different for each Talos version.

drivers/gasket/vars.yaml
```diff
@@ -0,0 +1 @@
+VERSION: "{{ .BUILD_ARG_TAG }}"
```
ammmze, line 1: I have mixed feelings about what to do for the version. I see in other places we have the library/module/app/etc version as a var and are using that (sometimes with this build arg tag). The issues I have with that are:

1. The version we put in the vars may or may not actually match the version of the module in the pkg. It seems like we would be better off extracting the version from somewhere in the pkg?
2. The gasket-driver doesn't actually have versions. It rarely changes, but the best we have is a commit sha.

Any thoughts or preferences here?

drivers/gasket/manifest.yaml
frezbo: this is fine, since the module would support any talos versions matching this constraint, the signature is not something that would prevent the module from installing in the first place

drivers/gasket/vars.yaml
frezbo: > The version we put in the vars may or may not actually match the version of the module in the pkg. It seems like we would > be better off extracting the version from somewhere in the pkg?

it depends, sometimes it might be completely impossible to infer that from `pkgs` so care has to be taken care while updating pkgs version. https://github.com/siderolabs/extensions/blob/main/Pkgfile#L8 we do the same here.

> The gasket-driver doesn't actually have versions. It rarely changes, but the best we have is a commit sha.

we could use the short commit sha

frezbo: I'll see if it's not too much work to extract this data out of pkgs.

ammmze:
FYI ... I just rebased this to resolve merge conflicts and I also added the driver to the extension catalog in the project root README

frezbo:
> FYI ... I just rebased this to resolve merge conflict (with the Makefile) and I also added the driver to the extension catalog in the project root README

Awesome, will get it merged before the 1.3.0 alpha talos release. Just have to do some cleanup on this repo first.

frezbo (approved):

smira (approved):

frezbo:
sorry for the holdup @ammmze

ammmze:
> sorry for the holdup @ammmze 

It's all good! Thanks for getting it merged!

PR #111, opened 2023-01-24
feat: add mellanox ofed drivers extension

frezbo:
Add Mellanox OFED drivers extension.
Keeping it disabled until it builds with kernel 6.1 Ref: https://github.com/siderolabs/pkgs/pull/661

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -51,6 +51,7 @@ TARGETS = \
 
 # Temporarily disabled, as drbd-pkg fails to build with Linux 6.1
 #		drbd \
+#       mellanox-ofed \
 
 NONFREE_TARGETS =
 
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -48,6 +48,7 @@ TARGETS = \
 		nvidia-container-toolkit \
 		nvidia-fabricmanager \
 		nvidia-open-gpu-kernel-modules
+# mellanox-ofed \
```
smira, line 51: move it down under the comment for drbd?

smira (approved):

Makefile
frezbo: fixed

frezbo:
> Can this extension please be (re)enabled?

I don't think it builds with current LTS kernel, community is free to propose patches or fix the build issues

PR #113, opened 2023-02-04
feat: add i915 microcode

DWSR:
Adds the microcode for the Graphics microcontroller (GuC) and H265/HEVC microcontroller (HuC). These binaries are required in order to enable low-power encoding, which is required for hardware transcoding on 9th and 11th generation Intel GPUs. More details are available here:

https://jellyfin.org/docs/general/administration/hardware-acceleration/#intel-gen9-and-gen11-igpus


Makefile
```diff
@@ -37,6 +37,7 @@ empty :=
 space = $(empty) $(empty)
 
 TARGETS = \
+		i915-ucode \
```
smira, line 40: small nit - please keep the list sorted

smira (approved):

frezbo (approved):

PR #138, opened 2023-03-30
fix: nvidia modules autoload

frezbo:
Fixes: https://github.com/siderolabs/talos/issues/7008

Prevent autloading of nvidia modules by default. This allows setting any extra kernel modules args from talos machineconfig.


smira (approved):
:+1:

PR #145, opened 2023-04-09
Introduce extension for rpi boot config loader.

OGKevin:
This introduces a new extension dedicated for Raspberrypi. The function of the extension is documented here:
https://github.com/OGKevin/talos-ext-rpi-generic#boot-configtxt-loader.

tl;dr
1. mount boot parition
2. override default config.txt file in that partition.

Closes https://github.com/siderolabs/talos/issues/5688


OGKevin:
@smira @frezbo can I get some feedback for this one?

platforms/generic-pi/boot-config-loader/pkg.yaml
```diff
@@ -0,0 +1,17 @@
+name: rpi-boot-config-loader
+variant: scratch
+shell: /toolchain/bin/bash
+dependencies:
+  - stage: base
+  - image: "ghcr.io/ogkevin/talos-ext-generic-pi:{{.VERSION}}"
```
frezbo, line 6: this seems like a private repo?

OGKevin: It's public 🤔 https://github.com/OGKevin/talos-ext-rpi-generic/pkgs/container/talos-ext-generic-pi

Unless I'm misunderstanding?

OGKevin: @frezbo little reminder poke 🙈

frezbo, line 6: quick nits:

we don't refer an external built tool, we download source and build since that means any security updates gets passed down the toolchain.

Secondly looking at the source, there seems to be the block devices are hardcoded as `/dev/mmcblk` etc, it would be easier to use the talos `go-blockdevice` library to find the parition with the `BOOT` label

OGKevin: RE your second point, I was planning on doing that indeed in a later stage. Not sure exactly how. Initial idea was to read the Talos config and get the install disk. Your suggestion also works. 

Is this feature required to get this MR merged? I can create an issue on the repo as well to track this progress if that helps.

OGKevin:
@frezbo do you mind 👀 again.

PR #154, opened 2023-04-28
Initial tailscale extension

btrepp:
This downloads and compiles tailscale,
as a talos extension, running as a service.

Uses the host/tun device.

Motivation is to enable/simplify networking, e.g by being in Tailscale, you can access your talos cluster anywhere, not needing to have routes setup if you are behind NAT devices, like would be common in hobby and lab setups.

```
_out
├── manifest.yaml
└── rootfs
    └── usr
        └── local
            ├── bin
            │   ├── tailscale
            │   └── tailscaled
            ├── etc
            │   └── containers
            │       └── tailscale.yaml
            └── lib
                └── containers
                    └── tailscale
                        └── containerboot
```

Installed like (I've only published arm at the moment)

### Installation.

```
machine:
  install:
    extensions:
      - image: docker.io/btrepp/tailscale:1.40.0-v1.4.0-alpha.4-1-g65239da
```

Install
``` talosctl upgrade```

Configure in Tailscale. It will log a message out, with a url to join the device.
I've chosen this approach as you don't need to manage a key, and would be 'interacting' with the operator on upgrades anyway, so if you nuke var you need to do this again.

1) `talosctl logs ext-tail scale`
2) find the login url and follow instructions
3) You would need to do this for each talos machine

### Persistence.

It mounts /var/lib/tailscale as rw, as Tailscale requires state to not create new nodes.
This could be an issue on upgrades, as if you don't use --preserve, you probably will have wiped the secrets Tailscale needs.

I've accepted this as an 'okay' tradeoff, in upgrades, you are manually firing commands `talosctl upgrade` anyway, so your procedure will just have a 'second' step, to register (and rename, if required) nodes in Tailscale.

The other option is ephemeral nodes and storing the auth key, which may be better, but Tailscale authkeys do expire, so it would be a trade-off.

### Addresses

Shows the addresses registered on the machine.
```
talosctl get address
NODE           NAMESPACE   TYPE            ID                                                       VERSION   ADDRESS                                       LINK
100.92.247.2   network     AddressStatus   eth0/192.168.1.11/24                                     1         192.168.1.11/24                               eth0
100.92.247.2   network     AddressStatus   eth0/2403:580f:43f:0:dea6:32ff:fedd:480d/64              1         2403:580f:43f:0:dea6:32ff:fedd:480d/64        eth0
100.92.247.2   network     AddressStatus   eth0/fe80::dea6:32ff:fedd:480d/64                        2         fe80::dea6:32ff:fedd:480d/64                  eth0
100.92.247.2   network     AddressStatus   lo/127.0.0.1/8                                           1         127.0.0.1/8                                   lo
100.92.247.2   network     AddressStatus   lo/::1/128                                               1         ::1/128                                       lo
100.92.247.2   network     AddressStatus   tailscale0/100.92.247.2/32                               1         100.92.247.2/32                               tailscale0
100.92.247.2   network     AddressStatus   tailscale0/fd7a:115c:a1e0:ab12:4843:cd96:625c:f702/128   1         fd7a:115c:a1e0:ab12:4843:cd96:625c:f702/128   tailscale0
100.92.247.2   network     AddressStatus   tailscale0/fe80::bb45:dc6d:c2ec:51e6/64                  1         fe80::bb45:dc6d:c2ec:51e6/64                  tailscale0
```

### CODE REVIEW

**Comment by btrepp** [comment]
There may be an issue as is, while talosctl seems happy enough, and you can use dns between nodes fine, k8s isn't up healthy. Kube-flannel seems to be struggling

```
kubectl logs -n kube-system kube-flannel-cc925
Defaulted container "kube-flannel" out of: kube-flannel, install-config (init), install-cni (init)
I0429 11:45:06.035373       1 main.go:211] CLI flags config: {etcdEndpoints:http://127.0.0.1:4001,http://127.0.0.1:2379 etcdPrefix:/coreos.com/network etcdKeyfile: etcdCertfile: etcdCAFile: etcdUsername: etcdPassword: version:false kubeSubnetMgr:true kubeApiUrl: kubeAnnotationPrefix:flannel.alpha.coreos.com kubeConfigFile: iface:[] ifaceRegex:[] ipMasq:true ifaceCanReach: subnetFile:/run/flannel/subnet.env publicIP: publicIPv6: subnetLeaseRenewMargin:60 healthzIP:0.0.0.0 healthzPort:0 iptablesResyncSeconds:5 iptablesForwardRules:true netConfPath:/etc/kube-flannel/net-conf.json setNodeNetworkUnavailable:true useMultiClusterCidr:false}
W0429 11:45:06.035645       1 client_config.go:617] Neither --kubeconfig nor --master was specified.  Using the inClusterConfig.  This might not work.
E0429 11:45:36.038711       1 main.go:228] Failed to create SubnetManager: error retrieving pod spec for 'kube-system/kube-flannel-cc925': Get "https://10.96.0.1:443/api/v1/namespaces/kube-system/pods/kube-flannel-cc925": dial tcp 10.96.0.1:443: i/o timeout
```

Unsure if that's related, but I suspect iptables changes for Tailscale may conflict somehow.

File: network/tailscale/tailscale.yaml [hunk 1]
```diff
@@ -0,0 +1,73 @@
...
+    options:
+      - bind
+      - rw
+  - source: /var/run/tailscale
+    destination: /var/run/tailscale
+    type: bind
+    options:
+      - bind
+      - rw
+  - source: /etc/resolv.conf
```
btrepp, line 67: This is potentially an issue.

Upgrading starts a race condition, where this points to the Tailscale service, which isn't running as it's being upgraded.

I thought this was meant to be more a 'ram' version in talos, but it seems modifying the file persists. Was hoping the 'reboot' procedure will go back to normal DNS resolution until the Tailscale service starts.

btrepp: --staged seemed to have helped. Unsure exactly why, but I think that does the bulk of the upgrade *after* a reboot, which may has reset to normal dns.

network/tailscale/tailscale.yaml
```diff
@@ -0,0 +1,73 @@
...
+  entrypoint: ./containerboot
+  args: []
+  environment:
+  - PATH=/usr/local/bin:/sbin
+  - TS_SOCKET=/var/run/tailscale/tailscaled.sock
+  - TS_STATE_DIR=/var/lib/tailscale
+  - TS_USERSPACE=false
+  - TS_ACCEPT_DNS=true
+  security:
+    maskedPaths: []
```
frezbo, line 18: do we need the `maskedPaths` and `readOnlyPaths` to be non defaults?

network/tailscale/tailscale.yaml
```diff
@@ -0,0 +1,73 @@
...
+    readonlyPaths: []
+    writeableRootfs: true
+    writeableSysfs: true
```
frezbo, line 21: I guess tailscale writes to some `sys` files, usually would only need access `/proc/sys` I beleive

network/tailscale/tailscale.yaml
```diff
@@ -0,0 +1,73 @@
...
+  environment:
+  - PATH=/usr/local/bin:/sbin
+  - TS_SOCKET=/var/run/tailscale/tailscaled.sock
+  - TS_STATE_DIR=/var/lib/tailscale
+  - TS_USERSPACE=false
+  - TS_ACCEPT_DNS=true
+  security:
+    maskedPaths: []
+    readonlyPaths: []
+    writeableRootfs: true
```
frezbo, line 20: since `bind` mounting `/var/lib/tailscale` this can be set to false

network/tailscale/tailscale.yaml
frezbo, line 67: containers by default gets the host resolv.conf, so does need to be bind mounted?

frezbo: oh, just saw that it's a `rw` mount, that seems tricky, since talos also has `/etc/resolv` as another bind mount. I guess a better option is to tell talos to update the dns entries, but that requires more work though

network/tailscale/vars.yaml
```diff
@@ -0,0 +1 @@
+VERSION: "{{ .TAILSCALE_VERSION }}-{{ .BUILD_ARG_TAG }}"
```
frezbo, line 1: i think just the tailscale version is fine. The `BUILD_ARG_TAG` is only used when the talos version has to match the extension version (eg: in case of kernel modules)

frezbo (commented):
Could you also add a README?

network/tailscale/tailscale.yaml
btrepp: Yeah I think that bind mount has caused me grief, so it's not the best. The main reason was to allow dns resolution. So that you can reference nodes via their hostnames, when connected.

It might be better to make that a user configurable option, e.g in your talos install, specify the Tailscale dns that the extension enables.

*but* that is also race condition, I imagine if the extension isn't up, you lose dns completely. It might be better to not have magicdns in the first part, though people would need to use the Tailscale IPs at the control plane layer.

I might think about this more, see if there is any alternative.

network/tailscale/tailscale.yaml
btrepp: I am not sure, it was part of debugging. I will try to pair it down

network/tailscale/tailscale.yaml
```diff
@@ -0,0 +1,73 @@
...
+    - etcfiles
+container:
+  entrypoint: ./containerboot
+  args: []
+  environment:
+  - PATH=/usr/local/bin:/sbin
+  - TS_SOCKET=/var/run/tailscale/tailscaled.sock
+  - TS_STATE_DIR=/var/lib/tailscale
+  - TS_USERSPACE=false
+  - TS_ACCEPT_DNS=true
```
btrepp, line 16: I think this maybe should be user configurable.
I might refactor the extension to read some configuration from a static file, including the TS_AUTH key so this would be more consistent over talos upgrades.

I want to do this before merging, as it's a significant change, so don't want someone to have a breaking change from this

frezbo: > I want to do this before merging, as it's a significant change, so don't want someone to have a breaking change from this

Cool, let us know when it's ready to be reviewed again, you could also mark the PR as a draft if it's still WIP

network/tailscale/tailscale.yaml
btrepp: Removed magic dns for first pass, it probably needs split dns as a seperate service, and is too much a PITA right now.

network/tailscale/tailscale.yaml
btrepp: Have removed it. Doesn't appear necessary

network/tailscale/tailscale.yaml
btrepp: set to false

network/tailscale/tailscale.yaml
btrepp: Completely removed resolve.conf binding. Magic DNS needs more design to integrate with talos

network/tailscale/vars.yaml
btrepp: Fixed

network/tailscale/tailscale.yaml
btrepp: Have changed this to use s6-envdir, so users can write MachineFiles into /var/etc/tailscale/env and set parts how they please

network/tailscale/pkg.yaml
```diff
@@ -0,0 +1,80 @@
...
+  - stage: base
+steps:
+  - env:
+      GOPATH: /go
+  - sources:
+      - url: https://github.com/tailscale/tailscale/archive/refs/tags/v{{ .TAILSCALE_VERSION }}.tar.gz
+        destination: tailscale.tar.gz
+        sha256: 6964176889943e0e0b25d8d69e14226cfb1c1a9944a257b24cb2dd212f797141
+        sha512: a64cad7685ba1dd438e16c8fa6ed1c29e5d840ffe17b2417342e43a2da5c1679826eb3a29b33e40ac075171a59605cdc857e1e09e60cf5140b22d363bb31ff30
+      - url: https://skarnet.org/software/s6/s6-{{ .S6_VERSION }}.tar.gz
```
frezbo, line 14: i would avoid adding all these dependencies and write  a simple wrapper like here: https://github.com/siderolabs/extensions/tree/main/storage/iscsi-tools/iscsid-wrapper

It just need to loop over files in a directory and set env vars from it

btrepp: Fair, it did feel getting excessive :). I don't know much go, and have used the s6 tools a lot before, so reached for what I know.

I'm also hoping any wrapper should also be short-lived, as if extensions supported envFiles, then it wouldn't be as necessary. I raised an issue for that. https://github.com/siderolabs/talos/issues/7316.

It may take me awhile to rewrite into a go wrapper, as I have no experience with go at the moment, and my system extension testing involves deploying to a RPI, so it's not the best turnaround, but I will give it a try

frezbo: @btrepp I've added the envfile support and cleaned up the PR. Sorry for the long wait.

frezbo:
@btrepp would it be possible to sign off the commit:

`git pull` to get the latest changes I pushed, then `git commit -s --amend --no-edit` and `git push --force-with-lease`

btrepp:
@frezbo should be done. Thanks for adding the environmentFile feature too!, that will be awesome for future extensions.

frezbo (approved):

PR #167, opened 2023-06-08
feat: added a new extension that installs the required drivers to operate usb modems under Talos

roeeklinger:
added a new extension that installs the required drivers to operate usb modems under Talos
requires: https://github.com/siderolabs/pkgs/pull/738


frezbo:
I can't push since the fork is from an org. I'll close and create a separate PR.

PR #173, opened 2023-06-21
Add `qemu-guest-agent` extension.

reitermarkus:
I haven't tested this at all yet since I haven't figured out yet what's the best way to do this.

cc @jabbas, since you mentioned this here https://github.com/siderolabs/talos/issues/3630#issuecomment-1399595011.

Try it using:

```yaml
machine:
  install:
    extensions:
      - image: docker.io/reitermarkus/qemu-guest-agent:8.0.2 
```

### DIFF
--- a/qemu/qemu-guest-agent/glib/pkg.yaml
+++ b/qemu/qemu-guest-agent/glib/pkg.yaml
@@ -1,0 +1,40 @@
+name: glib
+variant: scratch
+shell: /toolchain/bin/bash
+dependencies:
+  - stage: base
+  - image: "{{ .PKGS_PREFIX }}/openssl:{{ .BUILD_ARG_PKGS }}"
+  - stage: pcre2
+steps:
+  - sources:
+      - url: https://download.gnome.org/sources/glib/2.76/glib-{{ .GLIB_VERSION }}.tar.xz
+        destination: glib.tar.xz
+        sha256: c0be444e403d7c3184d1f394f89f0b644710b5e9331b54fa4e8b5037813ad32a
+        sha512: 291b8913918d411b679442b888f56893a857a77decfe428086c8bd1da1949498938ddb0bf254ed99d192e4a09b5e8cee1905fd6932ee642463fb229cac7c226e
+    prepare:
+      - |
+        tar -xf glib.tar.xz --strip-components=1
+        rm glib.tar.xz
+      - |
+        mkdir -p /usr/bin
+
+        ln -s /toolchain/bin/env /usr/bin/env
+        ln -s /toolchain/bin/python3 /toolchain/bin/python
+
+        pip3 install ninja
+    build:
+      - |
+        export PKG_CONFIG_PATH=/usr/lib/pkgconfig
+
+        meson setup \
+          --prefix=/usr \
+          -Ddefault_library=both \
+          _build
+
+        ninja -C _build
+    install:
+      - |
+        DESTDIR=/rootfs ninja -C _build install
+finalize:
+  - from: /rootfs
+    to: /
--- a/qemu/qemu-guest-agent/pcre2/pkg.yaml
+++ b/qemu/qemu-guest-agent/pcre2/pkg.yaml
@@ -1,0 +1,28 @@
+name: pcre2
+variant: scratch
+shell: /toolchain/bin/bash
+dependencies:
+  - stage: base
+steps:
+  - sources:
+      - url: https://github.com/PCRE2Project/pcre2/releases/download/pcre2-{{ .PCRE2_VERSION }}/pcre2-{{ .PCRE2_VERSION }}.tar.bz2
+        destination: pcre2.tar.bz2
+        sha256: 8d36cd8cb6ea2a4c2bb358ff6411b0c788633a2a45dabbf1aeb4b701d1b5e840
+        sha512: 72fbde87fecec3aa4b47225dd919ea1d55e97f2cbcf02aba26e5a0d3b1ffb58c25a80a9ef069eb99f9cf4e41ba9604ad06a7ec159870e1e875d86820e12256d3
+    prepare:
+      - |
+        tar -xf pcre2.tar.bz2 --strip-components=1
+        rm pcre2.tar.bz2
+    build:
+      - |
+        export PKG_CONFIG_PATH=/usr/lib/pkgconfig
+
+        ./configure --prefix=/usr
+
+        make -j $(nproc)
+    install:
+      - |
+        make DESTDIR=/rootfs install
+finalize:
+  - from: /rootfs
+    to: /
--- a/qemu/qemu-guest-agent/pkg.yaml
+++ b/qemu/qemu-guest-agent/pkg.yaml
@@ -1,0 +1,72 @@
+name: qemu-guest-agent
+variant: scratch
+shell: /toolchain/bin/bash
+dependencies:
+  - stage: base
+  - image: "{{ .PKGS_PREFIX }}/openssl:{{ .BUILD_ARG_PKGS }}"
+  - stage: pcre2
+  - stage: glib
+steps:
+  - sources:
+      - url: https://download.qemu.org/qemu-{{ .QEMU_VERSION }}.tar.xz
+        destination: qemu.tar.xz
+        sha256: f060abd435fbe6794125e2c398568ffc3cfa540042596907a8b18edca34cf6a5
+        sha512: 4e915d33a662bf55b09247fb85150be376c92270d3764e3d6470c452cb70cc558f54e84de5610dd60a9eb3ea02d5d4277b1ec75c9804967d278fa8361c7f9b9a
+    prepare:
+      - |
+        sed -i 's#$VERSION#{{ .VERSION }}#' /pkg/manifest.yaml
+      - |
+        tar -xf qemu.tar.xz --strip-components=1
+        rm qemu.tar.xz
+      - |
+        mkdir -p /usr/bin
+
+        ln -s /toolchain/bin/env /usr/bin/env
+        ln -s /toolchain/bin/python3 /toolchain/bin/python
+
+        pip3 install ninja
+    build:
+      - |
+        extra_args=( )
+
+        if [[ "${ARCH}" == aarch64 ]]; then
+          # FIXME: Workaround for “read-only segment has dynamic relocations” linking error.
+          extra_args+=( --disable-pie )
+        fi
+
+        export PKG_CONFIG_PATH=/usr/lib/pkgconfig
+
+        ./configure \
+          --static \
+          --prefix=/usr/local \
+          --localstatedir=/var \
+          --without-default-features \
+          --disable-system \
+          --disable-user \
+          --disable-docs \
+          --disable-install-blobs \
+          --enable-guest-agent \
+          --enable-stack-protector \
+          "${extra_args[@]}"
+
+        make -j $(nproc) qemu-ga
+    install:
+      - |
+        make DESTDIR=/rootfs install
+
+        container_root=/rootfs/usr/local/lib/containers/qemu-guest-agent
+        mkdir -p "${container_root}"
+
+        mv /rootfs/usr/local/bin/qemu-ga "${container_root}/qemu-ga"
+        rmdir /rootfs/usr/local/bin
+        rm -r /rootfs/usr/local/share/qemu
+        rmdir /rootfs/usr/local/share
+        rmdir /rootfs/var/run
+        rmdir /rootfs/var
+finalize:
+  - from: /rootfs
+    to: /rootfs
+  - from: /pkg/manifest.yaml
+    to: /
+  - from: /pkg/qemu-guest-agent.yaml
+    to: /rootfs/usr/local/etc/containers/
--- a/qemu/qemu-guest-agent/qemu-guest-agent.yaml
+++ b/qemu/qemu-guest-agent/qemu-guest-agent.yaml
@@ -1,0 +1,51 @@
+name: qemu-guest-agent
+depends:
+  - service: cri
+  - path: /dev/virtio-ports/org.qemu.guest_agent.0
+container:
+  entrypoint: ./qemu-ga
+  mounts:
+    # Shared libraries.
+    - source: /lib
+      destination: /lib
+      type: bind
+      options:
+        - bind
+        - ro
+    - source: /usr/lib
+      destination: /usr/lib
+      type: bind
+      options:
+        - bind
+        - ro
+    # State files.
+    - source: /var/run
+      destination: /var/run
+      type: bind
+      options:
+        - rshared
+        - rbind
+        - rw
+    # Device files.
+    - source: /dev
+      destination: /dev
+      type: bind
+      options:
+        - rshared
+        - rbind
+        - rw
+    # `/sbin/init` talks to `apid`.
+    - source: /system/run/apid/apid.sock
+      destination: /system/run/apid/apid.sock
+      type: bind
+      options:
+        - rshared
+        - rbind
+        - ro
+    - source: /sbin/init
+      destination: /sbin/shutdown
+      type: bind
+      options:
+        - bind
+        - ro
+restart: always
--- a/qemu/qemu-guest-agent/vars.yaml
+++ b/qemu/qemu-guest-agent/vars.yaml
@@ -1,0 +1,1 @@
+VERSION: "{{ .QEMU_VERSION }}"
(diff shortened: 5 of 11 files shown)


frezbo:
This should be fairly easy to test with `talosctl cluster create --provisioner=qemu --config-patch @<path to patch file containing extenions>`

reitermarkus:
@frezbo, I'm on macOS, so can't use `qemu` provisioner to test locally.

I tried applying the following patch to a node running in a Proxmox VM:

```
talosctl patch machineconfig --patch '[{"op": "add", "path": "/machine/install/extensions/-", "value": {"image": "docker.io/reitermarkus/qemu-guest-agent::8.0.2-v1.5.0-alpha.0-3-g988d89e"}}]' -n <node>
```

The service doesn't show up in `talosctl service -n <node>` though, even after a node reboot.

frezbo:
> The service doesn't show up in talosctl service -n <node> though, even after a node reboot.

you'd need to upgrade to have the extensions be picked up and installed (upgrading to same version is also fine)

reitermarkus:
Thanks, it is running now, or at least trying to:

```
exec ./qemu-ga: no such file or directory
```


`/rootfs/usr/local/lib/containers/qemu-guest-agent/qemu-ga` definitely exists though.

frezbo:
> 

it might not be a static binary or it's missing some shared libraries (which would show this error)

reitermarkus:
Thanks @frezbo, that might be it. Speaking of static compilation, do I need a statically compiled `bash`/`sh` in order to run a shell script in an extension?

frezbo:
> Thanks @frezbo, that might be it. Speaking of static compilation, do I need a statically compiled `bash`/`sh` in order to run a shell script in an extension?

not really, you'd have to then ship all the shared libraries it needs. We usually just write simple Go programs instead of scripts (you'd need so many deps to even get a shell :sweat_smile: )

reitermarkus:
Okay, I guess I can also patch the QEMU source code to call `/sbin/poweroff` instead of `/sbin/shutdown` instead of writing a shim script.

reitermarkus:
Would probably make sense for Talos to also provide `/sbin/shutdown` to support reboot via QEMU Guest Agent as well.

frezbo:
> Would probably make sense for Talos to also provide `/sbin/shutdown` to support reboot via QEMU Guest Agent as well.

i think that should be fine

qemu/qemu-guest-agent/pkg.yaml
```diff
@@ -0,0 +1,75 @@
+name: qemu-guest-agent
+variant: scratch
+shell: /bin/bash
+dependencies:
+  - image: debian:{{ .DEBIAN_IMAGE_VERSION }}
```
frezbo, line 5: this needs to use the siderolabs toolchain, only exception is that if the app does not work without glbc

reitermarkus:
@frezbo, I managed to build this using the toolchain, almost …

`linux/amd64` builds successfully, but `linux/arm64` currently fails with a linking error

```
#0 304.2 FAILED: qga/qemu-ga 
#0 304.2 cc  -o qga/qemu-ga qga/qemu-ga.p/meson-generated_.._qga-qapi-commands.c.o qga/qemu-ga.p/meson-generated_.._qga-qapi-emit-events.c.o qga/qemu-ga.p/meson-generated_.._qga-qapi-events.c.o qga/qemu-ga.p/meson-generated_.._qga-qapi-init-commands.c.o qga/qemu-ga.p/meson-generated_.._qga-qapi-introspect.c.o qga/qemu-ga.p/meson-generated_.._qga-qapi-types.c.o qga/qemu-ga.p/meson-generated_.._qga-qapi-visit.c.o qga/qemu-ga.p/commands.c.o qga/qemu-ga.p/guest-agent-command-state.c.o qga/qemu-ga.p/main.c.o qga/qemu-ga.p/cutils.c.o qga/qemu-ga.p/channel-posix.c.o qga/qemu-ga.p/commands-posix.c.o qga/qemu-ga.p/commands-posix-ssh.c.o qga/qemu-ga.p/commands-linux.c.o -Wl,--as-needed -Wl,--no-undefined -pie -Wl,--whole-archive -Wl,--start-group libevent-loop-base.a libqom.fa -Wl,--no-whole-archive -fstack-protector-strong -static-pie -Wl,-z,relro -Wl,-z,now -Wl,--warn-common -g0 -Os -s libqemuutil.a libqom.fa -lm -pthread -L/toolchain/lib -lgthread-2.0 -lglib-2.0 -lpcre2-8 -Wl,--end-group
#0 304.2 /toolchain/lib/gcc/aarch64-linux-musl/12.3.0/../../../../aarch64-linux-musl/bin/ld: read-only segment has dynamic relocations
#0 304.2 collect2: error: ld returned 1 exit status
#0 304.2 ninja: build stopped: subcommand failed.
#0 304.3 make[1]: *** [Makefile:165: run-ninja] Error 1
#0 304.3 make[1]: Leaving directory '/tmp/build/0/build'
#0 304.3 make: *** [GNUmakefile:11: qemu-ga] Error 2
```

reitermarkus:
Disabling PIE seems to avoid the linking error.

frezbo:
> Disabling PIE seems to avoid the linking error.

yes, that seems about right, (some apps don;t play well with gcc 12, we've had patches similar to disabling PIE)

power/nut-client/patches/replace_system.patch
```diff
@@ -14,19 +14,17 @@
  
 +static int runcmd(const char *cmd)
```
frezbo, line 15: i guess the new patch is from upstream?

qemu/qemu-guest-agent/glib/pkg.yaml
```diff
@@ -0,0 +1,38 @@
...
+        mkdir -p /usr/bin
+
+        ln -s /toolchain/bin/env /usr/bin/env
+        ln -s /toolchain/bin/python3 /toolchain/bin/python
+
+        pip3 install ninja
+    build:
+      - |
+        meson setup \
+          --prefix="${TOOLCHAIN}" \
```
frezbo, line 28: prefix shouldn't be toolchain, it shoud be `/usr`, since this is a runtime dependency

qemu/qemu-guest-agent/pcre2/pkg.yaml
```diff
@@ -0,0 +1,26 @@
...
+        destination: pcre2.tar.bz2
+        sha256: 8d36cd8cb6ea2a4c2bb358ff6411b0c788633a2a45dabbf1aeb4b701d1b5e840
+        sha512: 72fbde87fecec3aa4b47225dd919ea1d55e97f2cbcf02aba26e5a0d3b1ffb58c25a80a9ef069eb99f9cf4e41ba9604ad06a7ec159870e1e875d86820e12256d3
+    prepare:
+      - |
+        tar -xf pcre2.tar.bz2 --strip-components=1
+        rm pcre2.tar.bz2
+    build:
+      - |
+        ./configure --prefix="${TOOLCHAIN}"
```
frezbo, line 18: same here

power/nut-client/patches/replace_system.patch
reitermarkus: No, this isn't upstreamed. I just noticed the exit code does not match how `system` behaves. I can move this into a separate PR if needed.

qemu/qemu-guest-agent/glib/pkg.yaml
reitermarkus: Is it a runtime dependency if the executable is statically compiled anyways?

frezbo: that's correct, but if using the toolchain you can compile as dynamic and it should still work

qemu/qemu-guest-agent/qemu-guest-agent.yaml
```diff
@@ -0,0 +1,51 @@
...
+        - rbind
+        - rw
+    # `/sbin/init` talks to `apid`.
+    - source: /system/run/apid/apid.sock
+      destination: /system/run/apid/apid.sock
+      type: bind
+      options:
+        - rshared
+        - rbind
+        - rw
```
frezbo, line 44: this could be `ro` since socket files has no concept or read/write

qemu/qemu-guest-agent/vars.yaml
```diff
@@ -0,0 +1 @@
+VERSION: "{{ .QEMU_VERSION }}-{{ .BUILD_ARG_TAG }}"
```
frezbo, line 1:
```suggestion
VERSION: "{{ .QEMU_VERSION }}"
```

since this doesn't depend on talos kernel version

qemu/vars.yaml
```diff
@@ -0,0 +1,6 @@
+# renovate: datasource=github-releases extractVersion=^(?<version>.*)$ depName=qemu/qemu
```
frezbo, line 1: this should be git tag since qemu repository doesn't have releases

qemu/vars.yaml
```diff
@@ -0,0 +1,6 @@
...
+QEMU_VERSION: 8.0.2
+# renovate: datasource=git-tags depName=https://gitlab.gnome.org/GNOME/glib.git
+GLIB_VERSION: 2.76.3
+# renovate: datasource=github-releases extractVersion=^pcre-(?<version>.*)$ depName=PCRE2Project/pcre2
```
frezbo, line 5:
```suggestion
# renovate: datasource=github-releases extractVersion=^pcre2-(?<version>.*)$ depName=PCRE2Project/pcre2
```

qemu/qemu-guest-agent/vars.yaml
reitermarkus: Technically it does depend on the Talos version since a future version will hopefully provide `/sbin/shutdown` so the patch here can be removed, which would then be a breaking change for older Talos versions.

frezbo: Will get the `/sbin/shutdown` for 1.5, so should be good

reitermarkus: Great to hear, thanks.

qemu/qemu-guest-agent/qemu-guest-agent.yaml
reitermarkus: Interesting, do you not need write access to connect to a socket?

I tried it with `ro` and it does work, but it is quite surprising.

frezbo: > I tried it with ro and it does work, but it is quite surprising.

probably something to do with the fact that init talks to it, but in general file permissions does not makes sense for socket files (if it doesn;'t work with ro keep rw for now)

frezbo:
is this good to go? Probably also update readme to explain the use case, I guess for eg promox needs this? I'm not sure

frezbo:
Could you also sign-off the PR, `git pull` then `git commit -s --amend --no-edit` and `git push --force-with-lease`

reitermarkus:
Should I remove the patch here now or test again once 1.5.0 is released?

frezbo:
> Should I remove the patch here now or test again once 1.5.0 is released?

is there a quick way to test with a local qemu cluster?

reitermarkus:
> @frezbo, I'm on macOS, so can't use `qemu` provisioner to test locally.

I also tried to use `talosctl` to test “locally” on a Proxmox node, but that failed with:

```
validating CIDR and reserving IPs
generating PKI and tokens
the required flash image was not found in any of the expected paths for (["/usr/share/ovmf/OVMF.fd" "/usr/share/OVMF/OVMF.fd" "/usr/share/OVMF/OVMF_CODE_4M.fd" "/usr/share/OVMF/OVMF_CODE_4M.secboot.fd"]), please install it with the package manager or specify --extra-uefi-search-paths
```

frezbo:
I do see an error:

```bash
{"execute":"guest-shutdown"}    
{"error": {"class": "GenericError", "desc": "child process has failed to shutdown"}}
```

trying to shutdown, also the guest does not seem to log anything

reitermarkus:
Is that with or without the patch?

frezbo:
> Is that with or without the patch?

without the patch, I'm now trying with the patch

reitermarkus:
Did you mount `/sbin/shutdown` instead of `/sbin/poweroff` without the patch?

frezbo:
> Did you mount `/sbin/shutdown` instead of `/sbin/poweroff` without the patch?

okay, missed that, but doesn't seem to work with the patch also:

```bash
{"execute":"guest-ping"}
{"return": {}}
{"execute":"guest-shutdown"}
{"error": {"class": "GenericError", "desc": "child process has failed to shutdown"}}
```

ping and guest-info works, not the shutdown

frezbo:
```bash
{"execute":"guest-shutdown", "arguments":{"mode": "halt"}}
{"error": {"class": "GenericError", "desc": "mode is not supported on Talos: halt"}}


{"execute":"guest-shutdown", "arguments":{"mode":"poweroff"}}
{"error": {"class": "GenericError", "desc": "mode is invalid (valid values are: halt|powerdown|reboot"}}
{"execute":"guest-shutdown", "arguments":{"mode":"powerdown"}}
{"error": {"class": "GenericError", "desc": "child process has failed to shutdown"}}
```

so the patch does indeed show an invalid mode, but shutdown still doesn't work

reitermarkus:
Weird. Shutdown definitely works with the patch and a Proxmox VM.

frezbo:
> Weird. Shutdown definitely works with the patch and a Proxmox VM.

I was connecting to the guest agent socket and executing commands, other commands do work fine, not sure why this doesn't

reitermarkus:
Maybe try again with `rw` on the `apid` socket just to be sure? Otherwise no idea what could be the problem.

frezbo:
> Maybe try again with `rw` on the `apid` socket just to be sure? Otherwise no idea what could be the problem.

Already tried that, no change. Anyways `/sbin/shutdown` works as expected, so might remove the patch? It would have been nice to test this extension works as expected in CI, guess we cannot.

frezbo:
Oh found an interesting thing, doesn't seem to work on worker nodes, controlplane nodes work fine

frezbo:
> Oh found an interesting thing, doesn't seem to work on worker nodes, controlplane nodes work fine

could you test with a worker node on Proxmox?

reitermarkus:
I only tested with one node on Proxmox, which is in fact a worker node.

frezbo:
worker

```bash
❯ sudo socat - unix-connect:/home/frezbo/.talos/clusters/talos-default/talos-default-worker-1.sock
{"execute":"guest-ping"}
{"return": {}}
{"execute":"guest-shutdown"}
{"error": {"class": "GenericError", "desc": "child process has failed to shutdown"}}
```

controlplane

```bash
☸ admin@talos-default-274 (default) in talos on  main +4 [$!] is 📦 v0.0.1 via  v20.3.1 as 🧙 took 31s99ms 
❯ sudo socat - unix-connect:/home/frezbo/.talos/clusters/talos-default/talos-default-controlplane-1.sock
{"execute":"guest-ping"}
{"return": {}}
{"execute":"guest-shutdown"}
```

frezbo:
Adding commands to test:

References:
* https://wiki.qemu.org/Features/GuestAgent
* https://qemu.readthedocs.io/en/latest/interop/qemu-ga-ref.html#qapidoc-0

Talos patch:

```bash
diff --git a/pkg/provision/providers/qemu/launch.go b/pkg/provision/providers/qemu/launch.go
index 1cc97be86..c09e2969c 100644
--- a/pkg/provision/providers/qemu/launch.go
+++ b/pkg/provision/providers/qemu/launch.go
@@ -255,6 +255,10 @@ func launchVM(config *LaunchConfig) error {
 		"-no-reboot",
 		"-boot", fmt.Sprintf("order=%s,reboot-timeout=5000", bootOrder),
 		"-smbios", fmt.Sprintf("type=1,uuid=%s", config.NodeUUID),
+		"-chardev",
+		fmt.Sprintf("socket,path=%s/%s.sock,server=on,wait=off,id=qga0", config.StatePath, config.Hostname),
+		"-device", "virtio-serial",
+		"-device", "virtserialport,chardev=qga0,name=org.qemu.guest_agent.0",
 	}
 
 	for _, disk := range config.DiskPaths {
```

Cluster create command:

```bash
sudo -E _out/talosctl-linux-amd64 cluster create --provisioner=qemu $REGISTRY_MIRROR_FLAGS --controlplanes=1 --workers=1 --with-bootloader=true --config-patch @/tmp/patch.yaml
```

`/tmp/patch.yaml`

```yaml
machine:
  install:
    extensions:
      - image: 127.0.0.1:5010/frezbo/qemu-guest-agent:8.0.2
```


Testing:

```bash
sudo socat - unix-connect:/home/frezbo/.talos/clusters/talos-default/talos-default-controlplane-1.sock
{"execute":"guest-ping"}
{"return": {}}
{"execute":"guest-shutdown"}
```

frezbo:
I've removed the patches (test with `/sbin/shutdown` and cleaned minor nits

frezbo (approved):

frezbo:
I've removed the nut-client patch, since it broke the build. That can be a different PR if needed.

PR #178, opened 2023-07-04
chore: bump dependencies

smira:
Pkgs bump + open-iscsi 2.1.9.


storage/iscsi-tools/vars.yaml
```diff
@@ -1,5 +1,5 @@
 VERSION: v0.1.4
 # renovate: datasource=github-tags depName=open-iscsi/open-iscsi
-OPEN_ISCSI_VERSION: 2.1.8
+OPEN_ISCSI_VERSION: 2.1.9
```
frezbo, line 3: i think this broke the build last time, probably some patches

smira: yep, exactly, I will take a look into it later...

frezbo (approved):

PR #181, opened 2023-07-19
chore: use machined for shutdown calls

frezbo:
Use machined socket for `shutdown`, `poweroff` alias.

Depends on: https://github.com/siderolabs/talos/pull/7489

### DIFF
--- a/power/nut-client/manifest.yaml
+++ b/power/nut-client/manifest.yaml
@@ -7,4 +7,4 @@ metadata:
     This system extension provides the network-ups-tools upsmon service.
   compatibility:
     talos:
-      version: ">= v1.2.0"
+      version: ">= v1.5.0"
--- a/power/nut-client/nut-client.yaml
+++ b/power/nut-client/nut-client.yaml
@@ -5,6 +5,7 @@ depends:
     - addresses
     - connectivity
     - etcfiles
+  - path: /system/run/machined/machine.sock
 container:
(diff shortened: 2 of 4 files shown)


power/nut-client/manifest.yaml
```diff
@@ -7,4 +7,4 @@ metadata:
     This system extension provides the network-ups-tools upsmon service.
   compatibility:
     talos:
-      version: ">= v1.2.0"
+      version: ">= v1.5.0"
```
frezbo, line 10: we can bump this without breaking older nut-clients since it's versioned with `BUILD_ARG_TAG`

power/nut-client/nut-client.yaml
```diff
@@ -5,6 +5,7 @@ depends:
     - addresses
     - connectivity
     - etcfiles
+  - path: /system/run/machined/machine.sock
```
frezbo, line 8: also add a wait so extension starts after file is present, otherwise in rare situation extension service creates a folder with the mount source path

qemu/qemu-guest-agent/pkg.yaml
```diff
@@ -48,7 +47,8 @@ steps:
           --enable-guest-agent \
           --enable-stack-protector \
           "${extra_args[@]}"
-
+    build:
```
frezbo, line 50: just moving only build commands to the build stage

PR #185, opened 2023-07-22
feat: add 'zfs' package

kvaps:
ZFS is a powerful and flexible file system that ensures data integrity and offers features like snapshots and storage pooling.


storage/zfs/manifest.yaml
```diff
@@ -0,0 +1,10 @@
+version: v1alpha1
+metadata:
+  name: zfs
+  version: "$VERSION"
+  author: Andrei Kvapil
+  description: |
+    This system extension provides kernel module driver for ZFS built against a specific Talos version.
+  compatibility:
+    talos:
+      version: ">= v1.4.4"
```
frezbo, line 10: since this is only going to built for 1.5 onwards, I'll update this to 1.5.0

frezbo (approved):

Issue #171, opened 2023-06-13
[nvidia-container-toolkit=530.41.03-v1.12.1] [talos=v1.4.5] crash on container creation when NVIDIA_DRIVER_CAPABILITIES contains "graphics"


Kolpa:
I'm not 100% sure if this is the right place to report the issue, feel free to point me in the right direction if not :)

## My setup:
I am running a 2 node cluster based on talos 1.4.5 and i've enabled the nvidia container toolkit extension by following the [official talos tutorial](https://www.talos.dev/v1.4/talos-guides/configuration/nvidia-gpu-proprietary/#upgrading-talos-and-enabling-the-nvidia-modules-and-the-system-extension) (changing all mentions of talos version `v1.4.4` to `v1.4.5` and using version `v0.14.0` of the nvidia device plugin)

The nodes both contain a RTX 2080 GPU

## The issue:
All conventional cuda based container workloads i've tested run without any problems, `nvidia-smi` from inside of a container also reports the proper gpu information.

whenever i add `graphics` to the `NVIDIA_DRIVER_CAPABILITIES` env variable container creation fails with the following status:
```
failed to create containerd task: 
failed to create shim task: 
OCI runtime create failed: 
failed to create NVIDIA Container Runtime: 
failed to construct OCI spec modifier: 
failed to construct discoverer: 
failed to create mounts discoverer: 
failed to construct library locator: 
error loading ldcache: 
open /etc/ld.so.cache: no such file or directory: unknown
```

i am honestly at a loss here, if anyone has any input why this might be happening i would greatly appreciate any help

Kolpa:
I've also noticed the dev documentation for the extensions references a patch to the ldcache resolver file which i cannot find in the actual repository: https://github.com/siderolabs/extensions/blob/main/nvidia-gpu/nvidia-container-toolkit/DEVELOPMENT.md#patches

frezbo:
Some patches got merged upstream, will take a look at this next week

maxpain:
Any updates? I have the same problem

maxpain:
Any ETA or possible workarounds?

smira:
You can do a fix, perform some testing and submit a PR to speed up things. 

Otherwise we'll get to it as soon as we can, no ETA is available atm.

maxpain:
`nvc_ldcache.c.patch` was deleted in this commit
https://github.com/siderolabs/extensions/commit/93a7a0df54362a4279eb5cf9fc83c1f3023e9648

[...]

PR #191, opened 2023-08-01
feat: update extension spec allowlist for opengl

frezbo:
NVIDIA OpenGL/Vulkan files needs to be explicitly in the new locations added.

Ref: https://github.com/NVIDIA/nvidia-container-toolkit/blob/v1.13.5/internal/discover/graphics.go#L83-L89

```bash
root@2f51980391de:/tmp/NVIDIA-Linux-x86_64-535.54.03# find /usr /etc -type f -name "*.json"
/usr/share/glvnd/egl_vendor.d/10_nvidia.json
/usr/share/egl/egl_external_platform.d/10_nvidia_wayland.json
/usr/share/egl/egl_external_platform.d/15_nvidia_gbm.json
/usr/lib/nvidia/egl_dummy_vendor.json
/etc/vulkan/icd.d/nvidia_icd.json
/etc/vulkan/implicit_layer.d/nvidia_layers.json
```

`egl_dummy_vendor.json` can be ignored.

Partial fix for #171

### DIFF
--- a/README.md
+++ b/README.md
@@ -142,6 +142,9 @@ Any paths in the `rootfs` should be contained within the following hierarchies:
 - `/lib64/ld-linux-x86-64.so.2`
 - `/usr/etc/udev/rules.d/`
 - `/usr/local/`
+- `/usr/share/glvnd/`
+- `/usr/share/egl/`
+- `/etc/vulkan/`
 
 ## Dependency Diagram
 


smira (commented):
I feel we should also update code?

frezbo:
> I feel we should also update code?

yes, I'm testing it out to make sure everything's in place

smira (approved):

PR #197, opened 2023-08-11
feat: add ZFS toolings

donch:
* Add ZFS tools.
* Add a service to automatically import ZFS pools on startup.

This extensions was built and tested on Talos 1.5.0
We have sucessfully tested it on Kubernetes with [OpenEBS](https://github.com/openebs/zfs-localpv)
The zpool-import-wrapper service works well, and ZFS pools are automatically imported on startup

If you want to test it and create a ZFS pool, you can use a node-shell with the host filesystem mounted on /host, 
Then within the node-shell :
```
mount -o bind /host/dev /dev
mount -o bind /host/usr/local/sbin /usr/local/sbin
```
You will be able to create a ZFS pool:
```
zpool create zfs-pool /dev/sdX
```

Let me know if you need further informations. 
Thanks


JJGadgets:
+1, lack of ZFS tools on host is also blocking [https://github.com/democratic-csi/democratic-csi](democratic-csi)'s zfs-local modes from working.

democratic-csi uses `chroot` for the ZFS tools expecting them to be available on the host, to offload the burden of maintaining tools' version sync with the kernel module to the host who should already manage the kernel module versioning.

travisghansen:
Not a lot of experience with talos but author democratic-csi and get questions about this a bit. For the startup procedure should any further tooling be included to support or work with native zfs encryption?

frezbo (approved):

storage/zfs/zfs-tools/pkg.yaml
```diff
@@ -0,0 +1,61 @@
...
+        make -j $(nproc)
+    install:
+      - |
+        make install-exec DESTDIR=/rootfs/tmpdir
+        make install-data DESTDIR=/rootfs/tmpdir
+        mkdir -p /rootfs/usr/local/{sbin,bin}
+        mkdir -p /rootfs/usr/etc/udev/rules.d/
+        cp -R /rootfs/tmpdir/usr/etc/udev/rules.d/* /rootfs/usr/etc/udev/rules.d
+        sed -i s,/usr/etc/udev/,/usr/local/sbin/,g /rootfs/usr/etc/udev/rules.d/*
+        mv /rootfs/tmpdir/usr/etc/udev/zvol_id /rootfs/usr/local/sbin/zvol_id
```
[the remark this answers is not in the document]
rdegez: Thanks for reviewing the patch!
I'm a bit puzzled when you say `these udev rules are not going to work since there is no shell in talos` and removed those.
We added those for a reason ;-)
Upon zvol creation in the zpool, those udev rules are creating symlink in /dev (and need the using the zvol_id binary for that matter, that's also why we put it there): It's required for openebs-zfs-localpv to work! (when the zvol is created, the CSI will try to format & mount the zpool block device using the created symlink.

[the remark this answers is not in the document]
frezbo: This was a wrong analysis, `zvol_id` is a binary, added the rules back based on slack thread.

storage/zfs/zpool-importer.yaml
```diff
@@ -0,0 +1,50 @@
...
+  entrypoint: /usr/local/sbin/zpool-importer
+  mounts:
+    - source: /dev
+      destination: /dev
+      type: bind
+      options:
+        - rshared
+        - rbind
+        - rw
+    # ld-musl-x86_64.so.1
```
smira, line 16: for my understanding, if the extension anyway execs in pid 1 mount namespace, why do we need all the mounts?

frezbo: ahh good point, I'll drop those and test

frezbo: ahh it still needs those:

```bash
10.5.0.3: 2023/08/21 12:13:28 zpool-importer: reexec failed: gons/reexec: ReexecAction.Run: child failed with stderr message "err: exit status 1, /dev/zfs and /proc/self/mounts are required.\nTry running 'udevadm trigger' and 'mount -t proc proc /proc' as root.\n"
```

I think i starts from the parent and targets the mount of pid 1

smira: hmmm... I'm not sure I understand this...

frezbo: okay `/dev` is not needed, the other mounts are shared libraries, from the looks of it it seems the shared libs are needed in the parent mount namespace

smira (approved):

PR #205, opened 2023-08-17
feat: add device-mapper extension

ndbrew:
Add a new extension for device-mapper kernel drivers


ndbrew:
related:  pkgs#791

frezbo:
This is not needed since it's included in Talos

PR #213, opened 2023-08-24
feat: add intel ice firmware

frezbo:
Add firmware for Intel ICE (E810).


smira (approved):
cool, we can remove from Talos initramfs!

PR #215, opened 2023-08-31
[WIP] feat: Added rook-ceph kernel module support package

DreamingRaven:
Following on from https://github.com/siderolabs/talos/issues/7687.

Certain modules are currently unavailable for [rook-ceph](https://rook.io/docs/rook/v1.12/Getting-Started/Prerequisites/prerequisites/) namely `nbd`. Instead of only creating an extension for nbd, I created a general extension to enable rook-ceph required module support to make it more portable in future rather than relying on in-built talos kernel defaults.

However I have struck two problems, the first is I am uncertain how to enable the kernel modules to package in the linux config using the [bldr tool](https://github.com/siderolabs/bldr).

Secondly the makefile for extensions appears to heavily rely on only docker specifically making it especially difficult to test on a cri-o + podman setup.

I was taking heavy inspiration from the existing drivers for zfs and usb-modem.

Any advice would be much appreciated!


dhess:
We would like this as well. See https://github.com/siderolabs/pkgs/pull/861

smira:
I'd guess we'll make it part of standard Talos `initramfs`, so no need for an exteions. @dhess thanks for your PR

PR #216, opened 2023-09-01
feat: add minimal util-linux

frezbo:
Add a minimal util-linus tools. Only `fstrim` is enabled for now.


smira (commented):
does it install partially same libraries that already exist on the host?

frezbo:
> does it install partially same libraries that already exist on the host?

yes, I could drop `/lib` from the extension, but it's kind of hard like have to keep track of the files extension brings and the same ones present already on host.

smira:
> > does it install partially same libraries that already exist on the host?
> 
> yes, I could drop `/lib` from the extension, but it's kind of hard like have to keep track of the files extension brings and the same ones present already on host.

I thought we already install `util-linux` libraries, I'm more concerned about potential conflicts in the future with host/extension having same libraries

frezbo:
> > > does it install partially same libraries that already exist on the host?
> > 
> > 
> > yes, I could drop `/lib` from the extension, but it's kind of hard like have to keep track of the files extension brings and the same ones present already on host.
> 
> I thought we already install `util-linux` libraries, I'm more concerned about potential conflicts in the future with host/extension having same libraries

Okay, I've dropped all the lib that talos already ships (which was the same) and verified it works. I'll add a test suite to talos extensions suite.

smira (approved):

PR #218, opened 2023-09-01
feat: add xe-guest-utilities extension

casdr:
This adds a service extension for xe-guest-utilities. This daemon communicates with the Xen host to provide network and OS information.


virtualization/xe-guest-utilities/xe-guest-utilities.yaml
```diff
@@ -0,0 +1,17 @@
...
+container:
+  entrypoint: /usr/local/bin/xe-daemon
+  environment:
+  - PATH=/sbin:/usr/local/bin
+  security:
+    writeableRootfs: false
+    writeableSysfs: true
+  mounts:
+  # libs
+  - source: /lib
```
frezbo, line 11: what's using `/lib`?

frezbo: @casdr is this PR good to go?

casdr: @frezbo sorry for the late response, xe-daemon uses /lib/modules. I'll make this mount more explicit.

frezbo: I also guess this need a proper Xen host to test?

casdr: @frezbo I don't think there's another way to test this unfortunately.

frezbo: no worries, I will exclude this from the talos extensions test suite

frezbo:
I'll do some minor cleanups, like moving both qemu and xen guest agents into a single folder, otherwise good to go. Thanks :heart:

casdr:
@frezbo go ahead! Thank you too!

virtualization/xe-guest-utilities/xe-guest-utilities.yaml
```diff
@@ -0,0 +1,17 @@
+name: xe-guest-utilities
+container:
+  entrypoint: /usr/local/bin/xe-daemon
+  environment:
```
frezbo, line 4: this is probably not need as `/sbin` should be already in PATH

casdr: @frezbo sure! Send me the url and I'll test it for you.

The other binaries are utilities which can't/won't be used, so these aren't required. `xe-linux-distribution` might be interesting to include, but doesn't have support for Talos at the moment. I will look into adding this.

The [udev rule](https://github.com/xenserver/xe-guest-utilities/blob/master/mk/xen-vcpu-hotplug.rules) uses a shell, which won't work on Talos.

Currently the xe-guest-utilities functionality is limited mostly to having support for guest power actions. We're working on implementing more features (like network information) by removing the need of a shell.

frezbo: I'll sent an image link soon

frezbo: `ghcr.io/frezbo/xe-guest-utilities:8.3.0`

casdr: `xenstore` is also a CLI utility to interact with the store. This isn't required by `xe-daemon`.

I'm not the maintainer of xe-guest-utilities, but see what I can do :) I'll test your image ASAP.

casdr:
It turns out the following mount is required as well:

```
    - source: /dev/xen
      destination: /dev/xen
      type: bind
      options:
        - bind
        - rw
```

`xe-daemon` will communicate through `/dev/xen/xenbus`

frezbo:
> It turns out the following mount is required as well:
> 
> ```
>     - source: /dev/xen
>       destination: /dev/xen
>       type: bind
>       options:
>         - bind
>         - rw
> ```
> 
> `xe-daemon` will communicate through `/dev/xen/xenbus`

okay, can you try this image then: `ghcr.io/frezbo/xe-guest-utilities:sha256@028a45a9fc22d270b13845ed819d10e4fc10311220b316fabeccd9672a0ab402`

casdr:
Awesome, that works!

> Management agent detected

frezbo:
> Awesome, that works!
> 
> > Management agent detected

perfect, pushing my changes

frezbo (approved):

PR #224, opened 2023-09-11
chore: update pkgs

smira:
This is mostly to test buildkit 0.12.


Unix4ever (approved):

frezbo (approved):

smira:
@frezbo util-linux SHAs were wrong (cross-checked with pkgs)

PR #225, opened 2023-09-11
chore: add extensions catalog

smira:
Push the list of all images built for a specific version of extension to the `extensions` image. This way one can get a list of versions of extensions compatible with Talos version.

Update `sign-images` step to use `cosign` interactive flow, this is easier. Skip signing already signed images.


smira:
I retroactively added the `extensions` and updated image signatures for 1.5.0-1.5.2, will do same for 1.4.x later.

PR #228, opened 2023-09-13
feat: add binfmt_misc

sergelogvinov:
Add binfmt-misc kernel support.

https://github.com/siderolabs/pkgs/pull/804

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -42,6 +42,7 @@ space = $(empty) $(empty)
 TARGETS = \
 		amdgpu-firmware \
 		amd-ucode \
+		binfmt-misc \
 		bnx2-bnx2x \
 		btrfs \
 		chelsio-drivers \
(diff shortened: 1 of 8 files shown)


storage/binfmt-misc/pkg.yaml
```diff
@@ -0,0 +1,22 @@
+name: binfmt-misc
+variant: scratch
+shell: /toolchain/bin/bash
+dependencies:
+  - stage: base
+  # The pkgs version for a particular release of Talos as defined in
+  # https://github.com/siderolabs/talos/blob/<talos version>/pkg/machinery/gendata/data/pkgs
+  - image: "{{ .PKGS_PREFIX }}/binfmt-misc-pkg:{{ .BUILD_ARG_PKGS }}"
```
smira, line 8: can we make it pull from `kernel` pkg

Makefile
```diff
@@ -42,6 +42,7 @@ space = $(empty) $(empty)
 TARGETS = \
 		amdgpu-firmware \
 		amd-ucode \
+		binfmt-misc \
```
smira, line 45: we also need to add it to `README` please

storage/binfmt-misc/README.md
```diff
@@ -0,0 +1,17 @@
+# binfmt misc
```
smira, line 1: I don't think this is `storage` (I mean folder)

sergelogvinov: I am not sure the best place for it. May be we will create the folder like `modules`?

smira: I moved it to `misc`

frezbo (approved):

PR #229, opened 2023-09-13
feat: add stargz-snapshotter

andrewrynhard:
Adds the Stargz Snapshotter service


container-runtime/vars.yaml
```diff
@@ -1,2 +1,4 @@
 # renovate: datasource=github-tags extractVersion=^release-(?<version>.*)$ depName=google/gvisor
 GVISOR_VERSION: 20231214.0
+# renovate: datasource=github-tags extractVersion=^release-(?<version>.*)$ depName=containerd/stargz-snapshotter
```
frezbo, line 3:
```suggestion
# renovate: datasource=github-releases  depName=containerd/stargz-snapshotter
```

container-runtime/stargz-snapshotter/pkg.yaml
```diff
@@ -0,0 +1,48 @@
...
+        chmod +x /rootfs/usr/local/bin/ctr-remote
+finalize:
+  - from: /rootfs
+    to: /rootfs
+  - from: /pkg/manifest.yaml
+    to: /
+  - from: /pkg/stargz-snapshotter.part
+    to: /rootfs/etc/cri/conf.d/stargz-snapshotter.part
+  - from: /pkg/config.toml
+    to: /rootfs/usr/local/etc/containerd-stargz-grpc/config.toml
```
frezbo, line 48: we need to copy the service manifest too `/usr/loca/etc/containers`

container-runtime/stargz-snapshotter/pkg.yaml
```diff
@@ -0,0 +1,48 @@
...
+        make containerd-stargz-grpc
+
+        make ctr-remote
+    install:
+      - |
+        mkdir -p /rootfs/usr/local/bin
+
+        cd ${GOPATH}/src/github.com/containerd/stargz-snapshotter
+
+        cp ./out/containerd-stargz-grpc /rootfs/usr/local/bin/containerd-stargz-grpc
```
frezbo, line 35: this need to be in the container rootfs, or the service definiton should add a mount from `/usr/local/bin`

container-runtime/stargz-snapshotter/stargz-snapshotter.yaml
```diff
@@ -0,0 +1,21 @@
+name: stargz-snapshotter
+container:
+  entrypoint: ./containerd-stargz-grpc
```
smira, line 3: as this service is using `/var`, it should made dependent on `cri` service to guarantee it starts when `/var` is mounted and stopped before it gets unmounted

smira (commented):
also fyi, it has serious bugs which are not fixed: https://github.com/containerd/stargz-snapshotter/issues/215

smira:
and also https://github.com/dragonflyoss/image-service/issues/50#issuecomment-766781845

container-runtime/stargz-snapshotter/stargz-snapshotter.yaml
andrewrynhard: Looks like it still works with this dep.

frezbo: I'm confused how the system containerd behaves in this case, since both cri and containerd uses the same config

rsmitty (approved):

smira (commented):
hm. this was merged without updating top-level `README.md`. I think I had review comment about it

andrewrynhard:
> hm. this was merged without updating top-level `README.md`. I think I had review comment about it

I don't see a comment with that but I can submit a PR to update.

PR #253, opened 2023-10-31
chore: move to gh actions

frezbo:
Move to using GitHub actions.


Makefile
```diff
@@ -77,7 +77,6 @@ NONFREE_TARGETS = \
 
 all: $(TARGETS) ## Builds all known pkgs.
 
-.PHONY: nonfree
```
smira, line 80: not sure why this got removed, if we ever create path `nonfree`, this will stop working as a target

smira (approved):

Makefile
frezbo: this was not present in `pkgs` and we run `make` and `make nonfree` separately

Does it make sense to build `nonfree` when just `make` is invoked?

smira: it's not about `all` in fact, but it's about how `make` works. without `.PHONY`, if the file named `nonfree` exists, `make` will skip the target. `.PHONY` says that this target actually doesn't produce a file named `nonfree`.

frezbo: ahh yes, i forgot that

PR #254, opened 2023-11-01
feat: update stargz extension

rsmitty:
This PR updates the stargz extension to pull in busybox, exec into the pid 1 mount namespace, then start the stargz snapshotter. This allows us to work around what seems to be mount propagation issues with extensions. Once those are better understood and fixed within Talos, this extenstion can hopefully be updated to remove the namespace switching.

### DIFF
--- a/container-runtime/stargz-snapshotter/pkg.yaml
+++ b/container-runtime/stargz-snapshotter/pkg.yaml
@@ -1,5 +1,5 @@
 name: stargz-snapshotter
-variant: scratch
+variant: alpine
 shell: /toolchain/bin/bash
 dependencies:
   - stage: base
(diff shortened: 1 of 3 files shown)


smira:
I don't think there is any issue with mount propagation in Talos, it was an issue with this extension in the first place. It has so many mounts that it's impossible for me to even see how they are getting layered.

rsmitty:
I'll let @frezbo chime in on that, because I think he also mentioned encountering something similar with zfs extension. Makes me less likely to think it's something specific to this extension.

frezbo:
Yeh, there is something missing on talos side

rsmitty:
Closing for now, as this is going to be fixed in a cleaner way on the back of a Talos PR for extensions being able to mount as shared

PR #257, opened 2023-11-02
chore: use kres to manage project

frezbo:
Move to using kres to manage project.


storage/iscsi-tools/open-iscsi/pkg.yaml
```diff
@@ -33,23 +34,27 @@ steps:
...
+
+        pip3 install ninja
     build:
       - |
         export PKG_CONFIG_PATH=/usr/lib/pkgconfig
 
-        # musl doesn't have glob_onlydir but by spec it's also not guaranteed to return only dirs anyway
+        mkdir -p output
+
         LDFLAGS="$LDFLAGS -L/usr/local/lib" \
```
frezbo, line 45: the `make` way of building was already deprecated for `iscsi-tools` for a while, bringing in gcc 13 broke the build, so switch to meson

smira (approved):

smira:
do we have reproducibility test as well?

frezbo:
> do we have reproducibility test as well?

we never had for extensions, should I add one?

smira:
> > do we have reproducibility test as well?
> 
> we never had for extensions, should I add one?

might be great to have it... less of an issue now, but we should be reproducible?

frezbo:
> > > do we have reproducibility test as well?
> > 
> > 
> > we never had for extensions, should I add one?
> 
> might be great to have it... less of an issue now, but we should be reproducible?

added reproducibility too, `SOURCE_DATE_EPOCH` was missing for some

PR #264, opened 2023-11-13
fix: enable rootfs propagation for stargz-snapshotter

smira:
See https://github.com/siderolabs/talos/pull/7953


frezbo (approved):

frezbo:
https://github.com/siderolabs/extensions/blob/main/container-runtime/stargz-snapshotter/manifest.yaml#L10 this needs updation, to use 1.6.0

smira:
> https://github.com/siderolabs/extensions/blob/main/container-runtime/stargz-snapshotter/manifest.yaml#L10 this needs updation, to use 1.6.0

thanks, good point, let me update

PR #265, opened 2023-11-13
fix: iscsi-tools install `prefix`

frezbo:
Fixes the `iscsi-tools` install prefix to be `/usr/local`. This broke when moving from `make` to `meson`.

### DIFF
--- a/storage/iscsi-tools/iscsid.yaml
+++ b/storage/iscsi-tools/iscsid.yaml
@@ -61,6 +61,13 @@ container:
       options:
         - bind
         - ro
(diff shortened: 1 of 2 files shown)


storage/iscsi-tools/iscsid.yaml
```diff
@@ -61,6 +61,13 @@ container:
       options:
         - bind
         - ro
+    - source: /var/lib/iscsi
```
frezbo, line 64: previously it used `/etc/iscsi` which was a mount

smira (approved):

PR #270, opened 2023-11-22
fix: extensions image push was missed

frezbo:
Push the `extensions` image, this was missed when the CI was moved from drone to GitHub actions.


smira (commented):
Should we run make extensions always, but skip pushing it?

frezbo:
> Should we run make extensions always, but skip pushing it?

doesn;t it fetch manifests from registry first, then it would fail on normal build, right?

smira:
> > Should we run make extensions always, but skip pushing it?
> 
> doesn;t it fetch manifests from registry first, then it would fail on normal build, right?

hmm... yes, true. probably you're right

smira (approved):

PR #271, opened 2023-11-22
feat: mdadm extension

sergelogvinov:
Add mdadm tool in root filesystem.


frezbo:
is there a way to test this from talos side, would love to add some tests here: https://github.com/siderolabs/talos/blob/main/internal/integration/api/extensions_qemu.go already have a set of helpers and tests here, should be easy to add

sergelogvinov:
Ready to go:

```shell
# talosctl --talosconfig _cfgs/talosconfig --nodes 192.168.10.48 get extensions                                                                    
NODE            NAMESPACE   TYPE              ID                                                               VERSION   NAME    VERSION
192.168.10.48   runtime     ExtensionStatus   000.ghcr.io-sergelogvinov-mdadm-v4.2-v1.6.0-alpha.2-2-gc74d8e1   1         mdadm   v4.2-v1.6.0-alpha.2-2-gc74d8e1

# talosctl --talosconfig _cfgs/talosconfig --nodes 192.168.10.48 disks         
NODE            DEV        MODEL           SERIAL   TYPE   UUID   WWID   MODALIAS      NAME   SIZE    BUS_PATH                                                                                SUBSYSTEM          SYSTEM_DISK
192.168.10.48   /dev/sda   QEMU HARDDISK   -        SSD    -      -      scsi:t-0x00   -      34 GB   /pci0000:00/0000:00:1e.0/0000:05:01.0/0000:06:05.0/virtio1/host6/target6:0:0/6:0:0:0/   /sys/class/block   *
192.168.10.48   /dev/sdb   QEMU HARDDISK   -        SSD    -      -      scsi:t-0x00   -      11 GB   /pci0000:00/0000:00:1e.0/0000:05:01.0/0000:06:05.0/virtio1/host6/target6:0:0/6:0:0:2/   /sys/class/block   
192.168.10.48   /dev/sdc   QEMU HARDDISK   -        SSD    -      -      scsi:t-0x00   -      11 GB   /pci0000:00/0000:00:1e.0/0000:05:01.0/0000:06:05.0/virtio1/host6/target6:0:0/6:0:0:1/   /sys/class/block   

# talosctl --talosconfig _cfgs/talosconfig --nodes 192.168.10.48 mounts | grep '/dev/md'
192.168.10.48   /dev/md127p1                 10.67      0.11       10.56           1.03%          /var/data

# mdadm --detail /dev/md127
/dev/md127:
           Version : 0.90
     Creation Time : Wed Nov 22 18:21:41 2023
        Raid Level : raid1
        Array Size : 10485696 (10.00 GiB 10.74 GB)
     Used Dev Size : 10485696 (10.00 GiB 10.74 GB)
      Raid Devices : 2
     Total Devices : 2
   Preferred Minor : 127
       Persistence : Superblock is persistent

       Update Time : Thu Nov 23 19:25:57 2023
             State : clean 
    Active Devices : 2
   Working Devices : 2
    Failed Devices : 0
     Spare Devices : 0

Consistency Policy : resync

              UUID : 86345734:5d4de8db:5a848257:e2917ea5 (local to host controlplane-41)
            Events : 0.24

    Number   Major   Minor   RaidDevice State
       0       8       32        0      active sync   /dev/sdc
       1       8       16        1      active sync   /dev/sdb
```

sergelogvinov:
> is there a way to test this from talos side, would love to add some tests here: https://github.com/siderolabs/talos/blob/main/internal/integration/api/extensions_qemu.go already have a set of helpers and tests here, should be easy to add

The first idea that comes to my mind is to check for the binary existence.
All other checks can be done only with extra disk (two or more) Maybe loop devices helps...

frezbo:
> All other checks can be done only with extra disk (two or more) Maybe loop devices helps...

the extensions test mount up extra disks, if you look at zfs tests we run commands to create a pool and verify it's still mounted after a reboot, i think we should add something similar along the lines for mdadm too

storage/mdadm/files/udev-md-raid-arrays.rules
```diff
@@ -0,0 +1,36 @@
+# do not edit this file, it will be overwritten on update
+
+SUBSYSTEM!="block", GOTO="md_end"
```
frezbo, line 3: does these comes from upstream, then we can avoid putting it here and se the upstream ones as is

storage/vars.yaml
```diff
@@ -1,2 +1,3 @@
 # renovate: datasource=github-tags extractVersion=^release-(?<version>.*)$ depName=libfuse/libfuse
 FUSE3_VERSION: 3.16.1
+MDADM_VERSION: 4.2
```
frezbo, line 3: missing renovate field

storage/mdadm/pkg.yaml
```diff
@@ -0,0 +1,39 @@
+name: mdadm
+variant: scratch
+shell: /toolchain/bin/bash
+dependencies:
+  - stage: base
+  - image: "{{ .PKGS_PREFIX }}/eudev:{{ .BUILD_ARG_PKGS }}"
+  - image: "{{ .PKGS_PREFIX }}/util-linux:{{ .BUILD_ARG_PKGS }}"
```
frezbo, line 7: `util-linux` is already part of Talos, so we can drop it

frezbo: same with `eudev`

storage/mdadm/pkg.yaml
```diff
@@ -0,0 +1,39 @@
...
+    build:
+      - |
+        mkdir -p /run/mdadm
+
+        CXFLAGS=-D_LARGEFILE64_SOURCE \
+        make -j $(nproc) mdadm
+    install:
+      - |
+        mkdir -p /rootfs/usr/local/sbin /rootfs/usr/etc/udev/rules.d
+        cp mdadm /rootfs/usr/local/sbin/mdadm
```
frezbo, line 32: if `make install` installs the udev rules and the binary, let's use that

frezbo:
> > All other checks can be done only with extra disk (two or more) Maybe loop devices helps...
> 
> the extensions test mount up extra disks, if you look at zfs tests we run commands to create a pool and verify it's still mounted after a reboot, i think we should add something similar along the lines for mdadm too

Just saw the readme, i can add a test in Talos

storage/mdadm/pkg.yaml
sergelogvinov: They required by gcc, to build mdadm...

frezbo: nevermind, i thought this was the final stage

storage/mdadm/files/udev-md-raid-arrays.rules
sergelogvinov: We cannot use udev-rules as is... Most of them expect shell and systemd in the system.
I've remove unnecessary rules. We did the same in eudev module too...

frezbo: okay, makes sense

storage/mdadm/pkg.yaml
sergelogvinov: i do not think it is good idea -- check what it has inside https://packages.ubuntu.com/focal/amd64/mdadm/filelist
Talos does not need `mdmon` (systemd use it and cron pipelines)
We need only mdadm and two udev rules to activate the RAIDs.

frezbo: cool :+1:

frezbo:
I'll try building this and add a test to talos to test, then merge it

frezbo (approved):

PR #289, opened 2023-12-19
feat: generate a single extensions information file for all extensions

Unix4ever:
Get all extensions manifests from all subfolders and put them into a single file to help with the scenario when it's necessary to read all possible extensions information.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -1,6 +1,6 @@
 # THIS FILE WAS AUTOMATICALLY GENERATED, PLEASE DO NOT EDIT.
 #
-# Generated on 2023-12-12T23:07:05Z by kres latest.
+# Generated on 2023-12-19T16:28:49Z by kres latest.
 
 name: default
 concurrency:
@@ -90,6 +90,9 @@ jobs:
             })
 
             return resp.data.labels.map(label => label.name)
+      - name: check-dirty
+        run: |
+          make check-dirty
       - name: extensions
         if: github.event_name != 'pull_request'
         env:
--- a/.kres.yaml
+++ b/.kres.yaml
@@ -50,6 +50,8 @@ spec:
 kind: auto.CustomSteps
 spec:
   steps:
+    - name: check-dirty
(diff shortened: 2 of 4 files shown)


hack/locate-manifests.sh
```diff
@@ -0,0 +1,3 @@
+#!/usr/bin/bash
+
+find ./ -name "manifest.yaml" -exec echo "---" \; -exec cat {} \; > extensions.yaml
```
frezbo, line 3: maybe just inline in the makefile itself?

Unix4ever: yeah, I thought the same now :) Already changed.

.github/workflows/ci.yaml
```diff
@@ -90,6 +90,11 @@ jobs:
             })
 
             return resp.data.labels.map(label => label.name)
+      - name: check-dirty
+        env:
+          PUSH: "true"
```
smira, line 95: this looks scary a bit

.kres.yaml
```diff
@@ -96,6 +100,34 @@ spec:
...
+    enabled: true
+    phony: true
+    depends:
+      - $(ARTIFACTS)/bldr
+    script:
+      - |
+        @find ./ -name "manifest.yaml" -print0 | env LC_ALL=en_US sort -z | xargs -r0 -I{} sh -c 'echo "---\\n$$(cat {})"' > extensions.yaml
+---
+kind: custom.Step
+name: check-dirty
```
smira, line 115: I feel like `check-dirty` should be implemented as a base step in our Kres flow, as it would work for almost any project, if we hook it up after `generate` (or could be even the last step in the `generate` sequence).

Unix4ever: well, we can't really add it to the `generate` step, as it's either go or js generate steps.
While this particular project is neither of those.
So we need to add it to some common part for all, but after we detect the project type.

smira: yep, makes sense

smira (approved):

Issue #293, opened 2024-01-09
[qemu-guest-agent] Not launching in maintenance mode (1.6.1)


mlinares1998:
I'm trying to automate Talos deployments via Terraform on my Proxmox cluster.
To be able to provision the VM's correctly QEMU GA has to be started on boot so the Terraform Proxmox provider can get the VM's IP addresses.

This seems possible now since v1.6.0 using early extension initialization feature on maintenance mode. 
Unfortunately QEMU GA seems to not be working unlike Xen GA. It get's stuck on "Waiting for service "cri" to be registered" message.

![image](https://github.com/siderolabs/extensions/assets/15200008/88ac7a3b-4484-429a-994f-df279d5f362c)

I'm using an ISO bundled with QEMU GA by Image Factory with the following configuration.

```
Your image schematic ID is: f4524a31eb3f0cf8dc1449b8f43755d5ae4d35fc5c5b27eb79c8b8aca590bdfa
customization:
    systemExtensions:
        officialExtensions:
            - siderolabs/amd-ucode
            - siderolabs/intel-ucode
            - siderolabs/qemu-guest-agent
```

QEMU GA is correctly listed by talosctl

```
❯ talosctl -n 10.120.0.35 get extensions -i
NODE   NAMESPACE   TYPE              ID   VERSION   NAME               VERSION
       runtime     ExtensionStatus   0    1         amd-ucode          20231111
       runtime     ExtensionStatus   1    1         intel-ucode        20231114
       runtime     ExtensionStatus   2    1         qemu-guest-agent   8.1.3
       runtime     ExtensionStatus   3    1         schematic          f4524a31eb3f0cf8dc1449b8f43755d5ae4d35fc5c5b27eb79c8b8aca590bdfa
```

Thanks!

frezbo:
This is due to this https://github.com/siderolabs/extensions/blob/main/guest-agents/qemu-guest-agent/qemu-guest-agent.yaml#L23-L29

maybe we can move this to `/system/run` which is available in maintenance mode, the caveat being it gets blown away on next boot since it's ephemeral, but I guess it doesn't need to persist anything

mlinares1998:
Hi @frezbo. Thanks for your time!
I have tested building the extension locally with your suggestion, then adding it to an iso by Imager but got no avail.
It still gets stuck waiting for "cri" service.
[...]

PR #294, opened 2024-01-12
fix: qemu-guest-agent not starting in maintenance mode.

mlinares1998:
Allows qemu-ga to run under maintenance mode.

Resolves: #293

### DIFF
--- a/guest-agents/qemu-guest-agent/qemu-guest-agent.yaml
+++ b/guest-agents/qemu-guest-agent/qemu-guest-agent.yaml
@@ -1,6 +1,5 @@
 name: qemu-guest-agent
 depends:
-  - service: cri
   - path: /system/run/machined/machine.sock
   - path: /dev/virtio-ports/org.qemu.guest_agent.0
 container:
@@ -20,7 +19,7 @@ container:
         - bind
         - ro
     # State files.
-    - source: /var/run
+    - source: /system/run/qemu-guest-agent
       destination: /var/run
       type: bind
       options:


frezbo (approved):
:cool:

guest-agents/qemu-guest-agent/qemu-guest-agent.yaml
```diff
@@ -20,7 +19,7 @@ container:
         - bind
         - ro
     # State files.
-    - source: /var/run
+    - source: /system/run/qemu-guest-agent
```
smira, line 22: does it need any state to be actually mounted from the host? given it's tmpfs, it just preserves state across service restarts, not host restarts (?)

frezbo: i don't think so, just needs a writable `/var/run`

smira: I just wonder if we need support for `emptyDir` (or some equivalent of it) mounts...

frezbo: this might fit well into our volumes ideas

smira: well, in this case it's per-container (or even per-container run) temporary mount, we could do same thing as kubelet does - create some temporary directory under `/system`  or `/tmp`, mount it up into the container, and remove the directory when the container stops?

smira: on other hand, it probably makes sense to backport it _this way_ to 1.6.x, and we can work on `emptyDir` mounts next

frezbo: > on other hand, it probably makes sense to backport it this way to 1.6.x, and we can work on emptyDir mounts next

yeh, let's get this merged in as and once we have emptyDir support, let's update

PR #299, opened 2024-01-17
feat: add applesmc-t2 to support fan control

aarnaud:
Avoid over heating by allowing fan control on mac mini 2018 late


smira:
We usually try to avoid having any kernel modules which are not part of the official Linux tree.

Is this driver part of the tree in some later Linux release?

aarnaud:
It's part of kernel but the one included with the kernel is incompatible with T2 chip.
https://github.com/t2linux/
So It will be perfect for a labs cluster to be able to have the module as extension and avoid having forks with unofficial build.
Obviously, it's best-effort, but it allow some people to try Talos.

PR #310, opened 2024-02-01
fix: set `DATABASE_PATH` for fabricmanager

frezbo:
This fixes fabricmanager crashing.


smira (approved):

smira:
I guess we should backport it to v1.5 and v1.6?

frezbo:
> I guess we should backport it to v1.5 and v1.6?

yes, not sure about 1.5, that must be an old driver

smira:
> > I guess we should backport it to v1.5 and v1.6?
> 
> yes, not sure about 1.5, that must be an old driver

I think the problem was first reported in 1.5.x Talos.

jpuyobro:
Yes, the problem was reported using the nvidia version 535.54.03. And this version is proposed for talos 1.5

frezbo:
> Yes, the problem was reported using the nvidia version 535.54.03. And this version is proposed for talos 1.5

Thanks for confirming, I'll get it backported too.

PR #322, opened 2024-02-15
feat: realtek firmware

linuxmaniac:
fixes https://github.com/siderolabs/pkgs/issues/895


firmware/realtek/pkg.yaml
```diff
@@ -0,0 +1,23 @@
+name: bnx2-bnx2x
```
frezbo, line 1:
```suggestion
name: realtek-firmware
```

firmware/realtek/manifest.yaml
```diff
@@ -0,0 +1,10 @@
+version: v1alpha1
+metadata:
+  name: realtek
```
frezbo, line 3:
```suggestion
  name: realtek-firmware
```

linuxmaniac:
That was fast! Using realtek-firmware as suggested

smira:
it should also be added to `README.md`, and via `.kres.yaml` to the `Makefile`

linuxmaniac:
> it should also be added to `README.md`, and via `.kres.yaml` to the `Makefile`

done

frezbo (approved):

Issue #323, opened 2024-02-19
xe-guest-utlities not reporting IP


bandyguy:
I'm having issues with xe-guest-utilities reporting an IP. 

Talos `1.6.4`
XCP-NG `8.2.1`
Built iso from `ghcr.io/siderolabs/xe-guest-utilities:8.3.1@sha256:d8a0a61022f9237fe9b946ff27edb9bc92eb28c1ae6905855022520e62e4232e`
When I boot, I can see talos load xe-guest-utilities, and get an ip from dhcp.
Xen reports "Management agent detected"

I'm uncertain how to debug this further, or what other logs may be helpful.

bandyguy:
This looks like it is likely an issue with the xe-guest-utilities version with the XCP-NG version of the host. 

Related conversation: https://xcp-ng.org/forum/topic/4584/alpine-linux-from-hub-xe-daemon-collectos-error

j3rwin:
I think the networking part is still WIP https://github.com/siderolabs/extensions/pull/218#discussion_r1317558722

PR #335, opened 2024-03-08
feat: add the new rust-based xen-guest-agent as extension

j3rwin:
The existing xe-guest-utilities extension is based on [xenserver/xe-guest-utilities](https://github.com/xenserver/xe-guest-utilities/tree/master) which appears to be [abandoned](https://github.com/xenserver/xe-guest-utilities/issues/118) in favour of new the [rust-based guest agent](https://gitlab.com/xen-project/xen-guest-agent). Moreover the existing guest-utilities is largely a go-wrapper around a bunch of command line tools. Memory is the only metric that it seems to be reporting as it just parses the /proc/meminfo file, rest of the metrics are collected using command line tools that don't exist in Talos and so it fails silently.

This PR adds the new guest agent and it also fixes [issue#323](https://github.com/siderolabs/extensions/issues/323)


guest-agents/xen-guest-agent/pkg.yaml
```diff
@@ -0,0 +1,37 @@
+name: xen-guest-agent
+variant: alpine
+shell: /bin/sh
+install:
+  - cargo
```
frezbo, line 5: okay, i guess this is better than us maintaining rust in our toolchain (until we have a strong reason to do so, since it seemed to fail freqeuntly)

guest-agents/xen-guest-agent/pkg.yaml
```diff
@@ -0,0 +1,37 @@
...
+        sed -i 's#$VERSION#{{ .VERSION }}#' /pkg/manifest.yaml
+      - |
+        tar -xzvf xe-guest-agent.tar.gz --strip-components=1
+    build:
+      -  cargo build -F static --release
+    install:
+      - |
+        containerRoot=/rootfs/usr/local/lib/containers/xen-guest-agent
+        mkdir -p "$containerRoot"/usr/local/lib
+        mv /usr/lib/libgcc* "$containerRoot/usr/local/lib/"
```
frezbo, line 29: hmm rust doesn't produce fully static binaries?

guest-agents/vars.yaml
```diff
@@ -6,3 +6,5 @@ GLIB_VERSION: 2.79.2
 PCRE2_VERSION: 10.42
 # renovate: datasource=github-releases extractVersion=^v(?<version>.*)$ depName=xenserver/xe-guest-utilities
 XE_GUEST_UTILITIES_VERSION: 8.4.0
```
frezbo, line 8: should we remove the old one?

Makefile
```diff
@@ -84,6 +84,7 @@ TARGETS += util-linux-tools
 TARGETS += v4l-uvc-drivers
 TARGETS += wasmedge
 TARGETS += xe-guest-utilities
+TARGETS += xen-guest-agent
```
frezbo, line 87: this needs to be added in `.kres.yaml` and run a `make rekres` to regenerate the Makefile.

guest-agents/xen-guest-agent/pkg.yaml
j3rwin: Yeah, I couldn't build it using base. This seems better.

guest-agents/vars.yaml
j3rwin: I think so, will reduce the confusion for new users

Makefile
j3rwin: Ah, didn't read the first line. Fixed it

guest-agents/xen-guest-agent/pkg.yaml
frezbo: I think the cargo target needs to be `unknown-linux-gnu :thinking: for musl

j3rwin: It should be something like `{ARCH}-unkown-linux-musl` right? If I target gnu, wouldn't it fail when running in Talos?

https://rust-lang.github.io/rfcs/1721-crt-static.html this might help. I'm checking it now

frezbo: ah yes, `unknown-linux-musl` i believe

j3rwin: `+crt-static` worked. The binary is now standalone.

frezbo (approved):

smira (approved):

PR #344, opened 2024-03-18
feat: update Intel CPU ucode to 20240312

smira:
See
[RFDS](https://www.intel.com/content/www/us/en/developer/articles/technical/software-security-guidance/advisory-guidance/register-file-data-sampling.html).


frezbo (approved):
also update release.toml, or that can happen before release too

smira:
> also update release.toml, or that can happen before release too

thanks, let me do that

PR #355, opened 2024-04-07
feat: add spin wasm runtime

0xE282B0:
This commit adds the spin runtime, analogous to the wasmedge runtime, to use Talos Linux with spinkube.dev


smira:
@0xE282B0 thanks for your PR, it looks good. Could you please top-level `README.md` with this extension?

smira (approved):
NB: we should a test on Talos side for it

PR #369, opened 2024-04-24
docs: improve ExtensionServiceConfig docs

### LINKED ISSUE #363: Extensions docs need some help
Admit this is a beta feature that was just introduced, so some of this is probably already known.

https://github.com/siderolabs/extensions?tab=readme-ov-file#installing-extensions
https://github.com/siderolabs/extensions/blob/main/network/tailscale/README.md#subnet-routing
https://github.com/siderolabs/extensions/blob/main/network/tailscale/README.md#usage

I've only installed 3 extensions so far. I used the Image Factory, which none of these docs suggested or recommended, and it seems to be the best supported way to bundle extensions for your Talos image. That aside... after reading all of these, I was left with the impression that what I should do is put that ExtensionServiceConfig into a myconfig.yaml file and apply it with `talosctl apply -n controlplaneip -f myconfig` -- what seems to have worked, instead, is to append it to the `controlplane.yaml` instead, then do `apply` and `upgrade` on the resulting controlplane.yaml+myconfig.yaml.

Sort of. That almost worked, but the Talos node I wound up with no longer had any of my extensions installed. I ran the upgrade again with --image (specified repeating the same value as what `install.image` config I had mentioned in the `controlplane.yaml` file) - I think I probably do need to go back and read more upgrade docs, but maybe there's some thing that could be improved about this workflow. I am running single-node mode, this is not a HA or production cluster.

Still not sure if that was the correct procedure but I found `talosctl get extensionserviceconfig` seems to reveal the config applied successfully where I expected to find it (and I'm able to get into my tailscale tailnet service network via subnet router, so I guess it worked!) It seemed like this stuff all worked yesterday too, but today I came back after upgrading with `--image` and found it had rolled everything back to 1.6.5, maybe a problem of having mismatched CLI. (Did I mention I'm new here)

Is there a Talos for Beginners docs where I should start reading again, or similar docs where any of this stuff is covered in dead simple terms laid out flat for newbies, or is ExtensionServiceConfig still too new to expect that kind of hand-holding in documentation about how to use it (this feature seems to have been released in beta only days ago, in the very tip beta release which I am using now 🎉 )

Anyway, I bring this up, because running `talosctl apply myconfig.yaml` emitted some kind of message about applying config without rebooting, and I was left stranded with basically no idea that I should run `talosctl upgrade` now. I'm still not sure how I figured that out. At some point I think my node rebooted itself, went into some kind of failure state, and I was able to recover it in maintenance mode. I don't think that was supposed to happen. I guess there might be more than one way to load an ExtensionServiceConfig into the cluster, I heard about some kind of Talos-Kubernetes API bridge configuration where you can add Talos config directly through the Kubernetes API. I guess that's a secondary method, and `talosctl apply` is the main way to do it.

But through misusing `talosctl apply` in straightforward ways that seemed to be only doing exactly just what the docs suggested I should do, I was able to unfortunately take down my cluster and put the control plane into unauthenticated/maintenance mode, so maybe some additional guard rails are in order.

I would help with this but I'm not sure where to begin, just wanted to report on my findings and open to learn and review docs, (and maybe help someone else find their way through the maze in case they tripped over all the same things I did...)

mdallaire:
Improve the ExtensionServiceConfig docs for nut-client and tailscale.
- Explain that the ExtensionServiceConfig needs to be appended to the main config
- Provide example apply command
- Provide command to validate that the configuration is properly created

resolves #363


network/tailscale/README.md
```diff
@@ -21,11 +21,31 @@ environment:
...
+kind: ExtensionServiceConfig
+name: tailscale
+...
+```
+
+Then apply the full configuration to your node with
 ```bash
-> talosctl apply -n node myconfig.yaml
-> talosctl upgrade -n node
+talosctl apply -n mynode -f mynodeconfig.yaml
```
frezbo, line 39: i think we can just do `talosctl patch mc @patch.yaml`

smira: yes, the easiest way to add a multi-doc config is to patch your config; if the doc doesn't exist, it will be appended; if it already exists, it will be updated

mdallaire: I updated the PR with your feedback. Thanks.

frezbo (approved):

PR #383, opened 2024-05-12
feat: update spin extension to v0.14.1

rgl:


container-runtime/vars.yaml
```diff
@@ -7,6 +7,6 @@ CLOUD_PROVIDER_AWS_VERSION: v1.30.0
 # renovate: datasource=github-tags depName=containerd/runwasi
 WASMEDGE_VERSION: v0.3.0
 # renovate: datasource=github-tags depName=spinkube/containerd-shim-spin
-SPIN_VERSION: v0.13.1
+SPIN_VERSION: v0.14.1
```
frezbo, line 10: this needs updating checksums too https://github.com/siderolabs/extensions/blob/main/container-runtime/spin/pkg.yaml#L9

rgl: oh, indeed. check it now please.

frezbo (approved):

PR #400, opened 2024-06-02
added additional zfs services to support encrypted volumes

runningman84:
This should work fine but I am not sure if the dependencies work that way. Please review...

### DIFF
--- a/storage/zfs/zfs-key loader.yaml
+++ b/storage/zfs/zfs-key loader.yaml
@@ -1,0 +1,62 @@
+name: zfs-key-loader
+depends:
+  - service: udevd
+  - service: cri
+  - service: ext-zpool-importer
+  - path: /dev/zfs
+container:
+  security:
+    rootfsPropagation: shared
+  entrypoint: /usr/local/sbin/zfs
+  args:
+    - load-key
+    - -a
+  mounts:
+    # ld-musl-x86_64.so.1
+    - source: /lib
+      destination: /lib
+      type: bind
+      options:
+        - bind
+        - ro
+    # libcrypto.so and libc.so
+    - source: /usr/lib
+      destination: /usr/lib
+      type: bind
+      options:
+        - bind
+        - ro
+    - source: /usr/local/lib
+      destination: /usr/local/lib
+      type: bind
+      options:
+        - bind
+        - ro
+    - source: /usr/local/sbin
+      destination: /usr/local/sbin
+      type: bind
+      options:
+        - bind
+        - ro
+    - source: /dev
+      destination: /dev
+      type: bind
+      options:
+        - rshared
+        - rbind
+        - rw
+    - source: /proc
+      destination: /proc
+      type: bind
+      options:
+        - rshared
+        - rbind
+        - rw
+    - source: /var
+      destination: /var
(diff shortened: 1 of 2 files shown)


smira:
I would really love to land some support for mounting/volumes in Talos 1.7 to avoid such workaround if possible.

runningman84:
Okay if you bring native support for that in 1.7 I am totally fine… if not please consider merging this workaround.

runningman84:
@smira I played arround with zfs a bit. Zfs encryption in talos is tricky at the moment:

* if you mount an encrypted volume (zfs load-key -a; zfs mount -a) using a debug pod the files are not accessible in another pod
* having a mix of encrypted and unencrypted zfs datasets results in none of them being automounted anymore
* normally all datasets are automated if you use zfs but sometimes this does not happen, so you have to call zfs mount --a , this is also problematic because pods might write data to some local fs instead of the real zfs fs. This can result in data loos.

I think there should be some tests and documentation for zfs usage to make this usable without these issues,

rdegez:
Hi there!

I wonder if we could provide some help or human bandwidth to make progress and clear the path forward as much as possible on that matter ?
We really miss having at least basic at-rest ZFSencryption working on Talos nodes 😢 

@smira following on what you said in june :
> I would really love to land some support for mounting/volumes in Talos 1.7 to avoid such workaround if possible.

So I suppose in a post Talos 1.9 world, this means extending/enriching [VolumeConfig] (https://www.talos.dev/v1.9/reference/configuration/block/volumeconfig) but I guess this is no small task.. 😅 

Maybe the solution proposed in this PR would not be such a bad temporary workaround until you got this figured out in a generic way in the volume management system ? 

Or even a simpler variation of it, only issuing a `zfs load-key -a` immediately after zpool import and leaving everything else to the CSI (like openebs-zfs-localpv) that should be able to transparently create/mount zfs volumes within the pool without even having to know if the root pool is encrypted or not (since volumes created in an encrypted root pool inherit encryption config).

Also, I think this "only deal with the root pool" strategy should make all the issue raised by @runningman84 in https://github.com/siderolabs/extensions/pull/400#issuecomment-2158522646 void.

This key-loading step could be conditioned to the detection of encrypted zpool (easy to do with zfs and zpool binary).
Or even simply with a `zfs load-key -n` (performing a dry-run), see [doc](https://openzfs.github.io/openzfs-docs/man/master/8/zfs-load-key.8.html)

Maybe not be such a bad temporary workaround until you got this figured in a generic way in the volume management system ?

We could work on a new PR and stress-test this a bit if that make sense ?

jfroy:
I contributed a service for the zfs extension (https://github.com/siderolabs/extensions/pull/513) which is included in Talos 1.9. This service runs `zpool import -fal` to import pools. So if you set `keylocation` on a pool to a file path stored on a Talos volume, you should get automatic encrypted pool import at boot. I store my keys in `/var/zfs` and use Talos disk encryption to secure that.

smira:
We don't have any bandwidth at the moment to work on ZFS yet, the Volume Management work is still ongoing.

rdegez:
@jfroy woah I totally missed that! Super cool thanks a lot! :-)

PR #405, opened 2024-06-11
chore: use fedora mirror for glibc

frezbo:
The glibc FTP mirror is really unstable.


smira (approved):

smira:
I got Drone disabled, as I had it enabled for release-1.7

PR #411, opened 2024-06-12
chore: drop `/proc` mounts.

frezbo:
We don't need these proc mounts anymore, also having them `rshared` causes weird issues with volume discovery.


smira:
> also having them `rshared` causes weird issues with volume discovery.

it actually causes weird issues with core Talos, as each container restart layers more mounts on top of host `/proc`, and that causes a DoS after a few service restarts

smira (approved):

Issue #380, opened 2024-05-08
zfs breaks nvidia-container-toolkit


hansom82:
Both extensions uses shared library `libc.so`. If both present on target host then nvidia-device-plugin crashes with error:

```
Error: failed to create containerd task: failed to create shim task: OCI runtime create failed: runc create failed: unable to start container process: error during container init: error running hook #0: error running hook: exit status 127, stdout: , stderr: Auto-detected mode as 'legacy'
/usr/local/bin/nvidia-container-cli: error while loading shared libraries: /usr/local/glibc/lib/libc.so: invalid ELF header: unknown
```

#### Environment
- Talos version: v1.7.1
- Kubernetes version: 1.28.7
- Platform: metal (Proxmox and Bare Metal)

/fyi @kvaps

jahanson:
This has caused an issue for me as well
![image](https://github.com/siderolabs/extensions/assets/1548511/e605fe17-09d1-4cc3-9d80-ba7e06b91e2d)

The libc.so seems to be an ASCII file when most are symbolic links to their partner .so.# files.
![image](https://github.com/siderolabs/extensions/assets/1548511/bca4363a-2362-40af-b191-49c959ec1cb3)

frezbo:
This is interesting though, zfs should be linked against the musl libc, so i guess the zfs extensions ship some extra libc, which it shouldn't

frezbo:
This is due to both nvidia and zfs needing libtirpc and the zfs libtirpc is linked against musl while the nvidia one is linked against glibc. These two extensions are mutually incompatible with each other for now

rdegez:
As heavy zfs users and soon-to-be nvidia extension users as well we would be glad providing assistance to help solving this one.

@frezbo what do you think would be the best way to fix this then ?

Build and use glibc in the zfs extension build as well (based on https://github.com/siderolabs/extensions/blob/main/nvidia-gpu/nvidia-container-toolkit/glibc/pkg.yaml for instance) instead of musl ?

Or somehow trick the build process of https://github.com/siderolabs/extensions/blob/main/nvidia-gpu/nvidia-container-toolkit/nvidia-container-cli/libtirpc/pkg.yaml to use musl instead of the glibc ?

From the libtirpc source package, the autoconf snippet handling libc detection is located in in `libtirpc-1.3.4/config.guess` :
```
<...>
case $UNAME_SYSTEM in
Linux|GNU|GNU/*)
        LIBC=unknown

        set_cc_for_build
[...]

PR #421, opened 2024-06-24
fix: glibc search paths for nvidia

frezbo:
Set `glibc/lib` as first `rpath` for `nvidia-container-cli`. Also install nvidia libraries to `/usr/local/glibc/lib` so any musl libraries lives separately.

`nvidia-container-cli` explicitly sets an `RPATH` as `$ORIGIN/../$LIB` here: https://gitlab.com/nvidia/container-toolkit/libnvidia-container/-/blob/v1.14.6/Makefile?ref_type=tags#L183, this means `/usr/local/lib` would be searched first, since `zfs` and nvidia ship their own `libtirpc`, `nvidia-container-cli` first tries to use the `libtirpc` shippeed with `zfs` at `/usr/local/lib` instead of the one at `/usr/local/glibc/lib`. Fix this by setting an additional `RPATH` as `$ORIGIN/../glibc/$LIB`, so that libraries in `/usr/local/glibc/lib` have higher preference.

```bash
❯ scanelf -r _out/rootfs/rootfs/usr/local/bin/nvidia-container-cli
 TYPE   RPATH FILE
ET_DYN $ORIGIN/../glibc/$LIB:$ORIGIN/../$LIB _out/rootfs/rootfs/usr/local/bin/nvidia-container-cli
```

Properly fixes: #380

Fixes from #401 and #410 were not complete.

Manually tested by spinning up a NVIDIA worker in AWS.

### DIFF
--- a/nvidia-gpu/nvidia-container-toolkit/glibc/ld.so.conf
+++ b/nvidia-gpu/nvidia-container-toolkit/glibc/ld.so.conf
@@ -1,6 +1,1 @@
-# libc default configuration
-/usr/local/lib
-
 /usr/local/glibc/lib
-/usr/lib
-/lib
--- a/nvidia-gpu/nvidia-container-toolkit/nvidia-container-cli/pkg.yaml
+++ b/nvidia-gpu/nvidia-container-toolkit/nvidia-container-cli/pkg.yaml
@@ -51,7 +51,7 @@ steps:
         cd libnvidia-container
 
         # LDLIBS=-L/usr/local/glibc/lib is set so that libnvidia-container-cli libs which are hardcoded as -llibname and not using pkg-config
-        CPPFLAGS="-I/usr/local/glibc/include/tirpc" LDLIBS="-L/usr/local/glibc/lib -ltirpc -lelf -lseccomp" make
+        CPPFLAGS="-I/usr/local/glibc/include/tirpc" LDLIBS="-L/usr/local/glibc/lib -ltirpc -lelf -lseccomp" LDFLAGS='-Wl,--rpath=\$$ORIGIN/../glibc/\$$LIB' make
     install:
       - |
         mkdir -p /rootfs
--- a/nvidia-gpu/nvidia-container-toolkit/nvidia-pkgs/pkg.yaml
+++ b/nvidia-gpu/nvidia-container-toolkit/nvidia-pkgs/pkg.yaml
@@ -32,16 +32,12 @@ steps:
         bash nvidia.run --extract-only
     install:
       - |
-        mkdir -p /rootfs/usr/local \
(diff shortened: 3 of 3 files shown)


nvidia-gpu/nvidia-container-toolkit/glibc/ld.so.conf
```diff
@@ -1,6 +1 @@
-# libc default configuration
```
frezbo, line 1: This is to make sure, glibc doesn't try to load non-glibc libs

nvidia-gpu/nvidia-container-toolkit/nvidia-container-cli/pkg.yaml
```diff
@@ -51,7 +51,7 @@ steps:
         cd libnvidia-container
 
         # LDLIBS=-L/usr/local/glibc/lib is set so that libnvidia-container-cli libs which are hardcoded as -llibname and not using pkg-config
-        CPPFLAGS="-I/usr/local/glibc/include/tirpc" LDLIBS="-L/usr/local/glibc/lib -ltirpc -lelf -lseccomp" make
+        CPPFLAGS="-I/usr/local/glibc/include/tirpc" LDLIBS="-L/usr/local/glibc/lib -ltirpc -lelf -lseccomp" LDFLAGS='-Wl,--rpath=\$$ORIGIN/../glibc/\$$LIB' make
```
frezbo, line 54: This is the actual fix needed

nvidia-gpu/nvidia-container-toolkit/nvidia-pkgs/pkg.yaml
```diff
@@ -32,16 +32,13 @@ steps:
...
-          /rootfs/usr/local/lib/containers/nvidia-persistenced \
-          /rootfs/usr/local/etc/containers \
-          /rootfs/usr/etc/udev/rules.d
-
         cd NVIDIA-Linux-*
 
         ./nvidia-installer --silent \
           --opengl-prefix=/rootfs/usr/local \
           --utility-prefix=/rootfs/usr/local \
+          --utility-libdir=glibc/lib \
```
frezbo, line 40: this will only copy some libraries, not all hence the manual move below at https://github.com/siderolabs/extensions/pull/421/files#diff-d91dc320bde5625cb12535516b84cc3cd36e7ae24b320a40852b8f22a38299beR61

frezbo: for some reason `nvidia-container-cli` can't find libs stored under `/usr/local/lib` so we explicitly keep them under the custom location at `/usr/local/glibc/lib`

utkuozdemir (approved):
🆒

PR #427, opened 2024-07-14
Add dvb-si2168-firmware extension

samip5:
This is for adding the required firmware file(s).

Requires an kernel with dvb support enabled, which the default Talos kernel doesn't have.


Makefile
```diff
@@ -91,11 +91,12 @@ TARGETS += vmtoolsd-guest-agent
 TARGETS += wasmedge
 TARGETS += xen-guest-agent
 TARGETS += zfs
+TARGETS += dvb-si2168-firmware
```
smira, line 94: this file is auto-generated from `.kres.yaml` via `make rekres`

smira:
Let's put this on hold until there is a full story available with kernel drivers.

Makefile
```diff
@@ -91,11 +91,12 @@ TARGETS += vmtoolsd-guest-agent
...
 NONFREE_TARGETS = nonfree-kmod-nvidia
 
 # help menu
 
-export define HELP_MENU_HEADER
+define HELP_MENU_HEADER
```
samip5, line 99: My make would otherwise error out. I don't know if I'm the only one that has that issue with the way it was defined previously.

Makefile
samip5: After running said `mak rekres`, the makefile was changed to a format which makes make not work.

```
-define HELP_MENU_HEADER
+export define HELP_MENU_HEADER
 # Getting Started

 To build this project, you must have the following installed:
@@ -144,8 +143,6 @@ respectively.

 endef

-export HELP_MENU_HEADER
-
```


```
$ make rekres
Makefile:103: *** missing separator.  Stop.
```

samip5: Was running GNU Make 3.81, despite their being newer GNU Make 4.4.1.

samip5:
> Let's put this on hold until there is a full story available with kernel drivers.

Now it is: https://github.com/siderolabs/pkgs/pull/1089

firmware/dvb-si2168-firmware/pkg.yaml
```diff
@@ -0,0 +1,34 @@
...
+    prepare:
+      - |
+        sed -i 's#$VERSION#{{ .VERSION }}#' /pkg/manifest.yaml
+      - |
+        tar -xvf dvb-firmware.tar.gz
+        rm dvb-firmware.tar.gz
+    install:
+      - |
+        mkdir -p /rootfs/lib/firmware
+        cp dvb-firmware-0.0.51/firmware/dvb-demod-si2168-02.fw /rootfs/lib/firmware
```
frezbo, line 21: shouldn't this copy all firmware?

samip5: No because there's lot of non-needed firmware files included in the download.

samip5:
Superseaded by #538 as it includes firmware.

PR #430, opened 2024-07-18
feat(mdadm): configure mdadm.conf path for use with extensionconfig

thoro:
Add configuration (patch) so that mdadm reads the config file from /usr/local/etc/mdadm.conf instead of /etc/mdadm.conf


tzabbi:
What is the current state here?

frezbo:
this needs a rebase first

tzabbi:
Will something happen here or will it be rejected?

frezbo:
This will be fixed in 1.14

PR #436, opened 2024-07-31
feat(mei): add extension to provide Intel Management Engine drivers

e3b0c442:
Add an extension to provide the Intel Management Engine drivers that were introduced in https://github.com/siderolabs/pkgs/pull/997.

This should resolve https://github.com/siderolabs/pkgs/issues/700.

I do not know if I am missing any pieces here, in particular the Makefile rebuild, it wasn't clear how to accomplish that; I used other kernel module drivers as a template.


drivers/mei/pkg.yaml
```diff
@@ -0,0 +1,34 @@
+name: thunderbolt
```
frezbo, line 1: intended?

drivers/mei/README.md
```diff
@@ -0,0 +1,49 @@
...
+If you need Thunderbolt/USB4 networking,
+enable `thunderbolt_net` module as well.
+
+```yaml
+machine:
+	kernel:
+		modules:
+			- name: mei_hdcp
+			- name: mei-gsc
+			- name: mei-me
```
frezbo, line 28: seems these drivers have pci id's so i guess udev would automatically load them? :thinking:

drivers/mei/pkg.yaml
e3b0c442: No, thanks. I let myself get rushed this morning.

drivers/mei/README.md
e3b0c442: For the use case of Intel Arc, no further changes/configs were required, including dev; assuming the modules are available, the GPU is usable.

e3b0c442:
Rebased on main and squashed.

drivers/mei/README.md
frezbo: i meant there's no need to explicitly load them, since these modules have pci id's, it should get automatically loaded

e3b0c442: Yeah, same page, different words :)

e3b0c442: Oh, I misunderstood.

Let me run some tests and get back to you on that.

e3b0c442: You are correct. Docs updated.

frezbo:
@e3b0c442  on further checking all modules it seems some doesn't have PCI id's, could you confirm if they get auto-loaded (if those are dependencies it could get auto-loaded)?

* INTEL_MEI_GSC
* INTEL_MEI_HDCP
* INTEL_MEI_PXP
* INTEL_MEI_WDT

if not those needs to be added to `.machine.kernel.modules`

drivers/mei/pkg.yaml
```diff
@@ -0,0 +1,34 @@
...
+  - prepare:
+      - |
+        sed -i 's#$VERSION#{{ .VERSION }}#' /pkg/manifest.yaml
+  - install:
+      - |
+        export KERNELRELEASE=$(find /lib/modules -type d -name "*-talos" -exec basename {} \+)
+
+        mkdir -p /rootfs
+
+        xargs -a /pkg/files/modules.txt -I {} install -D /lib/modules/${KERNELRELEASE}/{} /rootfs/lib/modules/${KERNELRELEASE}/{}
```
smira, line 19: I guess we need to make this run for amd64 only, as this is not applicable (and we have no modules) for arm64 to make the extension "empty" for arm64

e3b0c442:
> @e3b0c442 on further checking all modules it seems some doesn't have PCI id's, could you confirm if they get auto-loaded (if those are dependencies it could get auto-loaded)?
> 
> * INTEL_MEI_GSC
> * INTEL_MEI_HDCP
> * INTEL_MEI_PXP
> * INTEL_MEI_WDT
> 
> if not those needs to be added to `.machine.kernel.modules`

In my testing I did confirm that all modules are listen in `/proc/modules` and the Intel GPU functionality I was depending on worked as expected, so I don't believe we need to explicitly load anything. Is there something else you'd like me to check specifically to verify?

drivers/mei/pkg.yaml
frezbo: but the step doesn't seem to fail, so I guess it's fine :thinking:

frezbo:
> > @e3b0c442 on further checking all modules it seems some doesn't have PCI id's, could you confirm if they get auto-loaded (if those are dependencies it could get auto-loaded)?
> > 
> > * INTEL_MEI_GSC
> > * INTEL_MEI_HDCP
> > * INTEL_MEI_PXP
> > * INTEL_MEI_WDT
> > 
> > if not those needs to be added to `.machine.kernel.modules`
> 
> In my testing I did confirm that all modules are listen in `/proc/modules` and the Intel GPU functionality I was depending on worked as expected, so I don't believe we need to explicitly load anything. Is there something else you'd like me to check specifically to verify?

awesome, thank you

drivers/mei/pkg.yaml
e3b0c442: is the CI pulling `pkgs` from `main`? If not I would actually expect both to fail since there hasn't been a `pkgs` release with the necessary modules.

frezbo: oh it's not, https://github.com/siderolabs/extensions/blob/main/.kres.yaml#L56 needs update here, weird it passed at all

frezbo: oh I know, `make rekres` was not run, so the targets was not added to `Makefile`

e3b0c442: OK, I ran that and took a swing at gating on arch. I don't have a complete understanding of the build architecture, so I made an educated guess that Docker's `TARGETARCH` would be available to the install script.

.conform.yaml
```diff
@@ -9,7 +9,7 @@ policies:
       gpg:
         required: true
         identity:
-          gitHubOrganization: siderolabs
+          gitHubOrganization: e3b0c442
```
frezbo, line 12: hmm, i can fix this

frezbo:
@e3b0c442 `mei_gsc_proxy` is not built it seems:

```bash
❯ crane export ghcr.io/siderolabs/kernel:v1.8.0-alpha.0-41-ga97d58f - | tar tf - | grep "mei"
lib/modules/6.6.43-talos/kernel/drivers/misc/mei
lib/modules/6.6.43-talos/kernel/drivers/misc/mei/hdcp
lib/modules/6.6.43-talos/kernel/drivers/misc/mei/hdcp/mei_hdcp.ko
lib/modules/6.6.43-talos/kernel/drivers/misc/mei/mei-gsc.ko
lib/modules/6.6.43-talos/kernel/drivers/misc/mei/mei-me.ko
lib/modules/6.6.43-talos/kernel/drivers/misc/mei/mei-txe.ko
lib/modules/6.6.43-talos/kernel/drivers/misc/mei/mei.ko
lib/modules/6.6.43-talos/kernel/drivers/misc/mei/pxp
lib/modules/6.6.43-talos/kernel/drivers/misc/mei/pxp/mei_pxp.ko
lib/modules/6.6.43-talos/kernel/drivers/watchdog/mei_wdt.ko
```

drivers/mei/pkg.yaml
frezbo: i'll fix this up

frezbo:
> @e3b0c442 `mei_gsc_proxy` is not built it seems:
> 
> ```shell
> ❯ crane export ghcr.io/siderolabs/kernel:v1.8.0-alpha.0-41-ga97d58f - | tar tf - | grep "mei"
> lib/modules/6.6.43-talos/kernel/drivers/misc/mei
> lib/modules/6.6.43-talos/kernel/drivers/misc/mei/hdcp
> lib/modules/6.6.43-talos/kernel/drivers/misc/mei/hdcp/mei_hdcp.ko
> lib/modules/6.6.43-talos/kernel/drivers/misc/mei/mei-gsc.ko
> lib/modules/6.6.43-talos/kernel/drivers/misc/mei/mei-me.ko
> lib/modules/6.6.43-talos/kernel/drivers/misc/mei/mei-txe.ko
> lib/modules/6.6.43-talos/kernel/drivers/misc/mei/mei.ko
> lib/modules/6.6.43-talos/kernel/drivers/misc/mei/pxp
> lib/modules/6.6.43-talos/kernel/drivers/misc/mei/pxp/mei_pxp.ko
> lib/modules/6.6.43-talos/kernel/drivers/watchdog/mei_wdt.ko
> ```


```text
2112:# CONFIG_INTEL_MEI_GSC_PROXY is not set
```

okay it's not enabled in pkgs, i'll drop it here

PR #438, opened 2024-07-31
feat: add crun container-runtime extension

hegerdes:
**About:**

This adds the [crun](https://github.com/containers/crun) container runtime-class to talos sytsem extentions.  
crun is a alternative implementation of the container-runtime-interface (cri) written in C.

It claims to 50% faster but the difference I conduced by a recent performance comparison between different cri's was 21% faster runc and it required less memory making it a great choice for low end and embedded systems.

**Tests:**
I created my own talos system image for the hetzner cloud (for amd64 & arm64) and created a cluster with this extension. With the RuntimeClass for crun applied I was able to create container just like the defualt runc cri.

For a quick test the vm-images assets in [this](https://github.com/hegerdes/helper-suite/releases/tag/v2024.07.01) GitHub release can be used to test the extention on hetzner.  
*Note:* The images were build bevor the changes regarding containerd config changes in ab77645a00cb074e2b52338540bca9a0cca72a6f and with the v1.7 PKGS

A pre-build docker image with the extentions file can be found [here](https://hub.docker.com/r/hegerdes/crun).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -64,6 +64,7 @@ TARGETS += drbd
 TARGETS += ecr-credential-provider
 TARGETS += fuse3
 TARGETS += gasket-driver
+TARGETS += crun
 TARGETS += gvisor
 TARGETS += gvisor-debug
 TARGETS += hello-world-service
--- a/container-runtime/crun/README.md
+++ b/container-runtime/crun/README.md
@@ -1,0 +1,43 @@
+# crun extension
+
+## Installation
+
+See [Installing Extensions](https://github.com/siderolabs/extensions#installing-extensions).
+
+## Usage
+
+In order to create the Wasm workload, a runtimeclass needs to be created.
+
+```yaml
+apiVersion: node.k8s.io/v1
+kind: RuntimeClass
+metadata:
+  name: crun
+handler: crun
+```
+
+## Testing
+
+Apply the following manifest to run nginx pod via crun:
+
+```yaml
+
+---
+apiVersion: v1
+kind: Pod
+metadata:
+  name: nginx-crun
+spec:
+  runtimeClassName: crun
+  containers:
+  - name: nginx
+    image: nginx
+```
+
+The pod should be up and running:
+
+```bash
+$ kubectl get pods
+NAME           READY   STATUS    RESTARTS   AGE
+nginx-crun   1/1     Running   0          40s
+```
--- a/container-runtime/crun/pkg.yaml
+++ b/container-runtime/crun/pkg.yaml
@@ -1,0 +1,40 @@
+name: crun
+variant: scratch
+shell: /toolchain/bin/bash
+dependencies:
+  - stage: base
+steps:
+  - sources:
+      # {{ if eq .ARCH "aarch64" }} This in fact is YAML comment, but Go templating instruction is evaluated by bldr
+      - url: https://github.com/containers/crun/releases/download/{{ .CRUN_VERSION }}/crun-{{ .CRUN_VERSION }}-linux-arm64-disable-systemd
+        destination: crun
+        sha256: 1bd840c95e9ae8edc25654dcf2481309724b9ff18ce95dbcd2535da9b026a47d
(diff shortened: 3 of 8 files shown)


container-runtime/crun/README.md
```diff
@@ -0,0 +1,46 @@
...
+The pod should be up and running:
+
+```bash
+$ kubectl get pods
+NAME           READY   STATUS    RESTARTS   AGE
+nginx-crun   1/1     Running   0          40s
+```
+
+
+git@github.com:hegerdes/talos-extensions.git
```
frezbo, line 46: remove this

hegerdes: sorry - missed that one in my cleanup step

Makefile
```diff
@@ -64,6 +64,7 @@ TARGETS += drbd
 TARGETS += ecr-credential-provider
 TARGETS += fuse3
 TARGETS += gasket-driver
+TARGETS += crun
```
smira, line 67: please add to the `.kres.yaml` and run `make rekres` to update `Makefile`

smira (commented):
please update top-level `README.md`

container-runtime/crun/pkg.yaml
```diff
@@ -0,0 +1,40 @@
+name: crun
+variant: scratch
+shell: /toolchain/bin/bash
+dependencies:
+  - stage: base
+steps:
+  - sources:
+      # {{ if eq .ARCH "aarch64" }} This in fact is YAML comment, but Go templating instruction is evaluated by bldr
+      - url: https://github.com/containers/crun/releases/download/{{ .CRUN_VERSION }}/crun-{{ .CRUN_VERSION }}-linux-arm64-disable-systemd
```
frezbo, line 9: we should also build from source probably

hegerdes: When is a requirement I can change to building from source. I'm already building crun from source with WASM enabled for another project so this should not be a problem. Just wanted to keep things simple.

Just one question: Do I need to handle any amd64/arm64 specific steps to build from source for both platforms or is the container build for both anyway an I can just assume I run on the right ARCH?

Version schema should follow talos upstream version. Is this correct?

hegerdes: I tried building it with the talos build environment. But I got some errors:

 * The crun release source tar.gz package does not contain the submodule code - so I need to clone the repo tag with submodules, annoying but not a problem.
 * Autogen does not find `libtoolize` - is there a way to install additional packages? Build ENV does seem to be based debian/alpine based.

Just for reference, on debian I install the following tools to install it:
```bash
apt-get install curl gcc build-essential pkgconf m4 libtool libsystemd-dev \
    libprotobuf-c-dev libcap-dev libseccomp-dev libyajl-dev libyajl2 \
    libprotobuf-c-dev protobuf-c-compiler protobuf-compiler libgcrypt20-dev \
    python3-protobuf go-md2man autoconf python3 automake \
    golang-github-checkpoint-restore-go-criu-dev
```

smira: this sounds like a lot of dependencies, probably let's skip building from source for now

frezbo: i think we have all the dependencies, i;ll take a look before re-signing, otherwise will fall back to binaries from release

frezbo: we're missing just yajl in our build environment, not sure if we add as a dependency and build or just use the static binaries

frezbo (approved):

hegerdes:
You guys are fast!
Was just in the middle of rebasing from main with the sugested changes for prebuild binary to bing the PR in a good state for now so I later can work on building from source.

Thanks for merging this (to fast) - you got a great product and a nice community ❤️

PS: Since I did not get to finish all your suggestions `crun` is currently missing in the top level Readme.md. Maybe you can add this later:

```markdown
| [crun](container-runtime/crun)                                       | [ghcr.io/siderolabs/crun](https://github.com/siderolabs/extensions/pkgs/container/crun)                                       | [crun](https://github.com/containers/crun) container runtime                                                                       | `upstream version` |
```

frezbo:
> PS: Since I did not get to finish all your suggestions crun is currently missing in the top level Readme.md. Maybe you can add this later:

I can fix this later

PR #445, opened 2024-08-06
feat: support lts and production nvidia modules

frezbo:
Support LTS and production versions of NVIDIA kernel modules as per https://docs.nvidia.com/datacenter/tesla/drivers/index.html#lifecycle

Part of: https://github.com/siderolabs/talos/issues/9086


smira (approved):
:sunglasses:

PR #454, opened 2024-08-17
feat: add openziti network extension

nenkoru:
Adds system extension to run OpenZiti edge tunneler which allows serving Talos nodes network resources over the overlay network.

> OpenZiti is a free and open source project focused on bringing zero trust networking principles directly into any application. The project provides all the pieces required to implement a zero trust overlay network and provides all the tools necessary to integrate zero trust into your existing solutions. The OpenZiti project believes the principles of zero trust shouldn't stop at your network, those ideas belong in your application.
https://openziti.io
https://github.com/openziti


nenkoru:
Idk how to solve gpg-identity EOF issue
I use mine ssh key to sign commits
Regarding number-of-commits - idk as well, branched from the main upstream.

frezbo:
> Idk how to solve gpg-identity EOF issue

you can ignore the gpg check, it checks if the commit is also signed by a member of org


> Regarding number-of-commits - idk as well, branched from the main upstream.

the checks is passing, so all good

network/openziti/pkg.yaml
```diff
@@ -0,0 +1,65 @@
...
+variant: alpine
+shell: /toolchain/bin/bash
+install:
+  - unzip
+  - zip
+  - ninja-build
+  - ninja
+  - zlib-static
+dependencies:
+  - image: "ghcr.io/siderolabs/tools:v1.7.0-4-gc844dc3"
```
smira, line 11: this looks a bit strange... we usually build from `base` layer which contains toolchain (tools) and don't use alpine for the build. is there anything missing?

nenkoru: There is no zlib-static, zip, unzip when using scratch and its impossible to use install directive as `apk` manager doesnt exist in `scratch' variant. 
Vcpkg relies on having zip, unzip as it downloads dependencies in that format. 
Zlib-static is required to link an executable against it as there is no zlib on the host /usr/lib folder

smira: zlib should be available though, zip and unzip, there will be no `apk` of course.

`vcpkg` downloads dependencies? does it pin them in a secure way?

network/openziti/pkg.yaml
```diff
@@ -0,0 +1,65 @@
...
+        # tried my best to find any CMAKE variable within vcpkg which disables this behaviour
+        # we fetch here commit to which tag 2024.07.12 is assigned
+        git clone https://github.com/microsoft/vcpkg.git /vcpkg
+        git --git-dir=/vcpkg/.git --work-tree=/vcpkg checkout 1de2026f28ead93ff1773e6e680387643e914ea1
+
+      - |
+        sed -i 's#$VERSION#{{ .VERSION }}#' /pkg/manifest.yaml
+
+      - |
+        mkdir -p /vcpkg-git /vcpkg /ziti-tunnel-sdk-c/build
```
frezbo, line 36: why are we building an sdk? doesn't make sense, shouldn't it be using this https://github.com/openziti/ziti/?

network/openziti/pkg.yaml
nenkoru: I guess I could just change to using 'base' stage instead of explicitly binding to a tools image here. 
Build seems to work fine with that. But idk what to do with those zip, unzip, zlib-static, ninja deps

network/openziti/pkg.yaml
nenkoru: > why are we building an sdk? doesn't make sense, shouldn't it be using this https://github.com/openziti/ziti/?

No we shouldn't. OpenZiti tunneler is exactly for hosting or tunneling. Main ziti repo is for the other part of the OpenZiti project which provides a controller, router and etc.

nenkoru: Tunneler is just a daemon that runs on the host machines and allows to put the network communication on or from the overlay network it creates. For Talos its impossible to change routing from the extension, so the tunneler runs in a 'run-host' mode which allows offloading of a overlay traffic on the underlay network.

network/openziti/pkg.yaml
nenkoru: > zlib should be available though, zip and unzip, there will be no `apk` of course.
> 
> `vcpkg` downloads dependencies? does it pin them in a secure way?

Yes it does by verifying their sha512sum as I skimmed the code.

nenkoru: > I guess I could just change to using 'base' stage instead of explicitly binding to a tools image here. Build seems to work fine with that. But idk what to do with those zip, unzip, zlib-static, ninja deps


Tried to build it with `- stage: base` and `variant: alpine`.
```
5.804 /toolchain/lib/gcc/x86_64-linux-musl/13.3.0/../../../../x86_64-linux-musl/bin/ld: attempted static link of dynamic object `/lib/libz.so'
5.804 collect2: error: ld returned 1 exit status
```
Which is obviously happening cuz of `-static`.

frezbo: seems there are static binaries available from the sdk releases, let's use them rather than adding all this build dependencies here, since it's going to be a lot of work to support it

nenkoru: > seems there are static binaries available from the sdk releases, let's use them rather than adding all this build dependencies here, since it's going to be a lot of work to support it

There are no static builds. 
```
$ file ziti-edge-tunnel (v1.1.1)
ziti-edge-tunnel: ELF 64-bit LSB pie executable, x86-64, version 1 (SYSV), dynamically linked, interpreter /lib64/ld-linux-x86-64.so.2, BuildID[sha1]=36243be4064e33f3c82f7a4e9767ffc3bed1e656, for GNU/Linux 3.2.0, with debug_info, not stripped
```

nenkoru: > seems there are static binaries available from the sdk releases, let's use them rather than adding all this build dependencies here, since it's going to be a lot of work to support it

Right, so what do you think next steps would be to get this merged upstream?

frezbo: hmm, then this extension needs to add it's build time dependencies as like qemu-guest-agent for example

nenkoru: > hmm, then this extension needs to add it's build time dependencies as like qemu-guest-agent for example

I guess would be hard to do as it heavily depends on vcpkg as the dependency manager. 
Maybe I could try working on creating static builds and then pulling them within the build process. Would this be okay?

nenkoru: I mean pulling them from the releases page, as you suggested above.

smira: yes, let's try static builds from the releases page. how big are those?

nenkoru: > yes, let's try static builds from the releases page. how big are those?

Well, around 5MB, I am currently working on buildin those using gh workflows, just having a linker issue with CMAKE trying to link libpthread.so instead of libpthread.a.
https://github.com/openziti/ziti-tunnel-sdk-c/issues/938

nenkoru:
> @nenkoru Do you have any news on this? Now openziti has static binaries.

Unfortunately, they are not

```bash
➜  Downloads file ziti-edge-tunnelt1712x64
ziti-edge-tunnelt1712x64: ELF 64-bit LSB pie executable, x86-64, version 1 (SYSV), dynamically linked, interpreter /lib64/ld-linux-x86-64.so.2, for GNU/Linux 3.2.0, BuildID[sha1]=8510e9ced615a6b8b9e790db3a23462c52424931, with debug_info, not stripped
```

Well, as far as I remember the team has decided to do not build static because of the issues they had with some devices and/or environments.
So for me I just proceed with rebuilding an extension every once in a while using the branch in the merge request and pushing it to private registry as well as my gh registry[1]. I have had some small tweaks over the past year, but the gist remains the same.
[1] https://github.com/nenkoru/talos-extensions/pkgs/container/openziti

PR #481, opened 2024-10-02
chore: mdadm, pkgs: update packages and replace eudev by systemd-udevd

dsseng:
Fix builds with new Talos packages versions

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,6 @@
 # THIS FILE WAS AUTOMATICALLY GENERATED, PLEASE DO NOT EDIT.
 #
-# Generated on 2024-09-18T14:25:18Z by kres 8be5fa7.
+# Generated on 2024-10-02T11:50:28Z by kres 8be5fa7.
 
 # common variables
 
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -48,7 +48,7 @@ COMMON_ARGS += --build-arg=PKGS_PREFIX="$(PKGS_PREFIX)"
 # extra variables
 
 EXTENSIONS_IMAGE_REF ?= $(REGISTRY_AND_USERNAME)/extensions:$(TAG)
-PKGS ?= v1.9.0-alpha.0-4-ge90ae7e
+PKGS ?= v1.9.0-alpha.0-11-g232a153
```
smira, line 51: this is generated, so we should update `.kres.yaml` and run `make rekres`

smira (commented):
plus I think between the pkgs version there were other bumps, at least Linux firmware, which should be updated in `Pkgfile`

Also we update `hack/release.toml` now to say that previous is 1.8.0, and remove old notes, leaving just new component updates - that is, Linux firmware at the very least

dsseng:
> plus I think between the pkgs version there were other bumps, at least Linux firmware, which should be updated in Pkgfile

None, the commit previously linked there was one where you updated the linux-firmware.

> Also we update hack/release.toml now to say that previous is 1.8.0, and remove old notes, leaving just new component updates - that is, Linux firmware at the very least

Should I do it in the same commit/PR?

smira:
> None, the commit previously linked there was one where you updated the linux-firmware.

ok, good then!

smira (approved):

Issue #482, opened 2024-10-02
Emulex Fibre Channel drivers


Hexoplon:
Currently, the Emulex FC adapter drivers are not available in the Talos image, or as an extension. Would be great to add the lpfc driver as an extension.

I have attempted to add it myself, but always run into issues with the modules kernel not being signed correctly.

smira:
As this is a storage driver, our policy is to bundle them with the base/default image, so we don't need an extension here.

PR #483, opened 2024-10-02
feat: add emulex fc hba driver extension

Hexoplon:
Add extension for enabling lpfc kernel module for Emulex Fibre Channel HBA modules.

#482
[Change in pkgs repo](https://github.com/siderolabs/pkgs/pull/1044)


smira:
We don't need an extension here, it will be part of base Talos.

PR #487, opened 2024-10-09
feat: introduce LLDPD extension service

### LINKED ISSUE #442: Add lldpd extension
This would be useful in bare-metal environments. I'm quite sad that I cannot figure out the switch ports each system is connected to via lldp without it.

It's C, so should be fairly simple compile to include: https://github.com/lldpd/lldpd

LLDP stands for Link Layer Discovery Protocol.

salkin:
- LLDPD service is depending on configuration to not start LLDPD before the user have ensured a LLDPD configuration they want to use is in place

fixes #442


frezbo:
Re-created as #492

frezbo:
> Re-created as #492

Since can't push to fork from organization

PR #488, opened 2024-10-11
Add pcirebind extension

mwdomino:
The `pcirebind` extension is a simple helper that attempts to rebind the drivers of a given PCI Bus ID during the boot process using a writable sysfs.

This is done by passing kernel args like:
```
machine:
  install:
    extensions:
      extraKernelArgs:
        - pcirebind.rebind=0000:04:00.00_ixgbe+vfio-pci
```

The above kernel args would generate actions similar to:
```
echo "vfio-pci" > /sys/bus/pci/devices/0000:04:00.0/driver_override
echo "0000:04:00.0" > /sys/bus/pci/drivers/ixgbe/unbind
echo "0000:04:00.0" > /sys/bus/pci/drivers/vfio-pci/bind
```

We're currently using this internally to rebind some NICs to use `vfio-pci` so that we can utilize them for VPP.

Let me know if you all have any questions, suggestions, or additions you'd like me to make on this!

### DIFF
--- a/misc/pcirebind/src/main.go
+++ b/misc/pcirebind/src/main.go
@@ -1,0 +1,154 @@
+package main
+
+import (
+	"fmt"
+	"os"
+	"strings"
+)
+
+type rebind struct {
+	id        string
+	oldDriver string
+	newDriver string
+}
+
+// writeToSysFile writes to the specified sysfs file
+func writeToSysFile(path, content string) error {
+	file, err := os.OpenFile(path, os.O_WRONLY, 0200) // 0200 is write-only permission
+	if err != nil {
+		e := fmt.Errorf("failed to open %s: %v", path, err)
+		fmt.Printf("[error] %v", e)
+
+		return e
+	}
+	defer file.Close()
+
+	if _, err = file.WriteString(content); err != nil {
+		e := fmt.Errorf("failed to write to %s: %v", path, err)
+		fmt.Printf("[error] %v", e)
+
+		return e
+	}
+
+	return nil
+}
+
+// parseKernelCmdline parses specially formatted strings out of /proc/cmdline
+//
+// The format is as follows:
+// pcirebind.rebind=<pci_bus_id>_<current_driver>+<new_driver>
+//
+// example:
+//
+//	pcirebind.rebind=0000:04:00.0_ixgbe+vfio-pci
+func parseKernelCmdline(readFunc func(string) ([]byte, error)) ([]rebind, error) {
+	data, err := readFunc("/proc/cmdline")
+	if err != nil {
+		return nil, fmt.Errorf("failed to read /proc/cmdline: %v", err)
+	}
+
+	cmdline := string(data)
+	parts := strings.Fields(cmdline)
+
+	var rebindsList []rebind
+
+	for _, part := range parts {
+		if strings.HasPrefix(part, "pcirebind.rebind=") {
+			// Extract the rebind specification
+			spec := strings.TrimPrefix(part, "pcirebind.rebind=")
+
+			// Split the specification into id, oldDriver, and newDriver
+			parts := strings.Split(spec, "_")
+			if len(parts) != 2 {
+				fmt.Printf("Invalid rebind format: %s\n", spec)
+				continue
+			}
+			id, drivers := parts[0], parts[1]
+
+			driverParts := strings.Split(drivers, "+")
+			if len(driverParts) != 2 {
+				fmt.Printf("Invalid drivers format: %s\n", drivers)
+				continue
+			}
+			oldDriver, newDriver := driverParts[0], driverParts[1]
+
+			// Append the parsed rebind information to the list
+			rebindsList = append(rebindsList, rebind{
+				id:        id,
+				oldDriver: oldDriver,
+				newDriver: newDriver,
+			})
+		}
+	}
+
+	if len(rebindsList) > 0 {
+		return rebindsList, nil
+	}
+
+	return nil, fmt.Errorf("no rebinds found in kernel command line")
+}
+
+// overrideDriver sets the `driver_override` for a given PCI Bus ID
+// this is analogous to:
+//
+//	echo "vfio-pci" > /sys/bus/pci/devices/0000:04:00.0/driver_override
+func (r *rebind) overrideDriver() error {
+	bindPath := fmt.Sprintf("/sys/bus/pci/devices/%s/driver_override", r.id)
+	return writeToSysFile(bindPath, r.newDriver)
+}
+
+// unbindDriver writes to `unbind` a given PCI Bus ID
+// this is analogous to:
+//
+// echo "0000:04:00.0" > /sys/bus/pci/drivers/ixgbe/unbind
+func (r *rebind) unbindDriver() error {
+	bindPath := fmt.Sprintf("/sys/bus/pci/drivers/%s/unbind", r.oldDriver)
+	return writeToSysFile(bindPath, r.id)
+}
+
+// bindDriver writes to `bind` a given PCI Bus ID
+// this is analogous to:
+//
+// echo "0000:04:00.0" > /sys/bus/pci/drivers/vfio-pci/bind
+func (r *rebind) bindDriver() error {
+	bindPath := fmt.Sprintf("/sys/bus/pci/drivers/%s/bind", r.newDriver)
+	return writeToSysFile(bindPath, r.id)
+}
+
+// v8 was last working version
+func main() {
+	rebinds, err := parseKernelCmdline(os.ReadFile)
+	if err != nil {
+		fmt.Printf("Error parsing kernel command line: %v\n", err)
+
+		os.Exit(1)
+	}
+
+	anyError := false
+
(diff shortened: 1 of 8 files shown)


mwdomino:
I'm not sure what any of those CI errors mean so just let me know if I need to make any changes to placate the bots :)

misc/pcirebind/src/main.go
```diff
@@ -0,0 +1,154 @@
...
+
+// parseKernelCmdline parses specially formatted strings out of /proc/cmdline
+//
+// The format is as follows:
+// pcirebind.rebind=<pci_bus_id>_<current_driver>+<new_driver>
+//
+// example:
+//
+//	pcirebind.rebind=0000:04:00.0_ixgbe+vfio-pci
+func parseKernelCmdline(readFunc func(string) ([]byte, error)) ([]rebind, error) {
```
frezbo, line 44: this could probably use `go-procfs` from here: https://github.com/siderolabs/go-procfs

frezbo: @smira i wonder if this needs to be an extension or core talos? :thinking:

smira:
I wonder if that should be a Talos controller which accepts the configuration like:

```yaml
kind: PCIRebindConfig
apiVersion: v1alpha1
name: 0000:04:00.00
drivers:
  old: ixgbe
  new: vfio-pci
```

frezbo:
While implementing a controller I came across conflicting/confusing docs about drivers binding and unbinding and not sure which to follow:

* There's a vfio specific implementation at kata-containers here: https://github.com/kata-containers/kata-containers/blob/main/src/runtime/pkg/device/drivers/vfio.go#L284
* There's a bash version of this which handles iommu too: https://github.com/andre-richter/vfio-pci-bind/blob/master/vfio-pci-bind.sh 

So for the above do we need to handle iommu? The PR uses the ` /sys/bus/pci/drivers/<driver_name>` path whereas the above two uses `/sys/bus/pci/devices/<vendorID>` path? Which one to prefer? Or does `unbind` from a driver means all iommu aalso gets unbinded?

I came across this https://code.google.com/archive/p/pci-hacking/wikis/bind_Uunbind_PCI.wiki which doesn't mention calling `driver_override` at all

and there's this kernel doc https://www.kernel.org/doc/Documentation/ABI/testing/sysfs-bus-pci which isn't much helpful

the only reference is an old LWN article which may not be relevant at all: https://lwn.net/Articles/143397/

@mwdomino any thoughts?

frezbo:
So the end goal from Talos side is to have a generic driver bind/unbind support rather than vfio-specific one

dsseng:
Some clarification, I haven't read the code fully yet.

> Writing to driver_override does not automatically unbind the
> device from its current driver or make any attempt to
> automatically load the specified driver.  If no driver with a
> matching name is currently loaded in the kernel, the device
> will not bind to any driver.

https://wiki.archlinux.org/title/PCI_passthrough_via_OVMF recommends using new_id to allow VFIO to bind to the device. Well, I read more and seems like driver_override enables us to rebind the device uniquely, which I believe is fairly important for Talos use case: https://wiki.archlinux.org/title/PCI_passthrough_via_OVMF#Using_identical_guest_and_host_GPUs . Here they also show that you have to walk around the IOMMU group and rebind all the devices except for the port because you can either passthrough all or none (will take another look).

Another thing they consider and we might want to either check or give user the ability to check would be: https://wiki.archlinux.org/title/PCI_passthrough_via_OVMF#Passing_through_a_device_that_does_not_support_resetting

> * There's a bash version of this which handles iommu too: https://github.com/andre-richter/vfio-pci-bind/blob/master/vfio-pci-bind.sh

Seems to just list and chown (I guess for the VMM to send some management to there) the IOMMU group.

misc/pcirebind/src/main.go
```diff
@@ -0,0 +1,154 @@
...
+}
+
+// bindDriver writes to `bind` a given PCI Bus ID
+// this is analogous to:
+//
+// echo "0000:04:00.0" > /sys/bus/pci/drivers/vfio-pci/bind
+func (r *rebind) bindDriver() error {
+	bindPath := fmt.Sprintf("/sys/bus/pci/drivers/%s/bind", r.newDriver)
+	return writeToSysFile(bindPath, r.id)
+}
```
dsseng, line 116: Take a look at [thread introducing driver_override](https://patchwork.kernel.org/project/linux-pci/patch/20140509164326.12141.44591.stgit@bling.home/#8907131). This allows us to just use `echo 0000:03:00.0 > /sys/bus/pci/drivers_probe` for kernel to determine the driver for the unbound device. Less parameters should be better I believe

misc/pcirebind/src/main.go
```diff
@@ -0,0 +1,154 @@
...
+
+// v8 was last working version
```
dsseng, line 118: unclear comment, please elaborate or remove it

dsseng:
> https://docs.kernel.org/driver-api/vfio.html the vfio docs also doesn't mention the `driver_override`

Yes, they add the ID to the `vfio-pci` driver. This is the older behavior, I believe it could cause unwanted binding of the driver in case e.g. we have one device, rebind it manually but we added its ID as wanted by VFIO, then another device with the same ID is added and it could get bound to VFIO as well. Using override on the exactly the needed device, then unbind, then bind or drivers_probe should clearly only touch the wanted device.

Everything else from VFIO docs is applicable. And I'm not sure whether or not we should rebind all the devices in the same IOMMU group

dsseng:
> > Everything else from VFIO docs is applicable. And I'm not sure whether or not we should rebind all the devices in the same IOMMU group
> 
> Yeh, that's something I'm confused about, i wonder if doing an unbind from `/sys/bus/pci/drivers/<driver_name>` unbinds all iommu also, since `/sys/bus/pci/devices/<vendorID>/iommu_group/devices/*` are basically symlinks back to drivers though 🤔

It shouldn't, I guess in most cases the IOMMU groups have been preallocated and not related to drivers themselves. Either on FW level or pcieport driver level.

Example from my laptop:

```
[    0.250230] [      T1] iommu: Default domain type: Passthrough
[    0.276575] [      T1] pci 0000:00:00.2: AMD-Vi: IOMMU performance counters supported
[    0.276622] [      T1] pci 0000:00:01.0: Adding to iommu group 0
[    0.276636] [      T1] pci 0000:00:02.0: Adding to iommu group 1
[    0.276643] [      T1] pci 0000:00:02.2: Adding to iommu group 2
[    0.276650] [      T1] pci 0000:00:02.4: Adding to iommu group 3
[    0.276663] [      T1] pci 0000:00:08.0: Adding to iommu group 4
[    0.276670] [      T1] pci 0000:00:08.1: Adding to iommu group 4
...
[    0.276806] [      T1] pci 0000:04:00.0: Adding to iommu group 4
[    0.276809] [      T1] pci 0000:04:00.1: Adding to iommu group 4
[    0.280354] [      T1] perf/amd_iommu: Detected AMD IOMMU #0 (2 banks, 4 counters/bank).
```

And amdgpu driver only loads on time or 4-5 second for this system, so it's the core PCI subsystem to log the group allocation.

mwdomino:
> While implementing a controller I came across conflicting/confusing docs about drivers binding and unbinding and not sure which to follow:
> 
>     * There's a vfio specific implementation at kata-containers here: https://github.com/kata-containers/kata-containers/blob/main/src/runtime/pkg/device/drivers/vfio.go#L284
> 
>     * There's a bash version of this which handles iommu too: https://github.com/andre-richter/vfio-pci-bind/blob/master/vfio-pci-bind.sh
> 
> 
> So for the above do we need to handle iommu? The PR uses the ` /sys/bus/pci/drivers/<driver_name>` path whereas the above two uses `/sys/bus/pci/devices/<vendorID>` path? Which one to prefer? Or does `unbind` from a driver means all iommu aalso gets unbinded?
> 
> I came across this https://code.google.com/archive/p/pci-hacking/wikis/bind_Uunbind_PCI.wiki which doesn't mention calling `driver_override` at all
> 
> and there's this kernel doc https://www.kernel.org/doc/Documentation/ABI/testing/sysfs-bus-pci which isn't much helpful
> 
> the only reference is an old LWN article which may not be relevant at all: https://lwn.net/Articles/143397/
> 
> @mwdomino any thoughts?

You all have done some awesome digging. I have to admit you seem to understand what is going on under the hood better than I do :) When I initially wrote this it was strictly for our internal use of swapping in the `vfio` driver for some Intel NICs so as long as we can still support this use-case anything that generalizes it better for other users' workloads seems like a great change.

mwdomino:
Yes. Our current kernel args for the POC we're working on are:

```yaml
        extraKernelArgs:
        - default_hugepagesz=2M
        - hugepages=2048
        - iommu=pt
        - intel_iommu=on
        - vfio_iommu_type1.allow_unsafe_interrupts=1
        - talos.network.interface.ignore=eno49
        - talos.network.interface.ignore=eno50
        - pcirebind.rebind=0000:04:00.0_ixgbe+vfio-pci
        - pcirebind.rebind=0000:04:00.1_ixgbe+vfio-pci
```

PR #501, opened 2024-10-21
chore: move udev rules

dsseng:
Ref: siderolabs/pkgs#1053

Signed-off-by: Dmitry Sharshakov <dmitry.sharshakov@siderolabs.com>

### DIFF
--- a/storage/iscsi-tools/open-iscsi/pkg.yaml
+++ b/storage/iscsi-tools/open-iscsi/pkg.yaml
@@ -51,7 +51,7 @@ steps:
           -Dno_systemd=true \
           -Dprefix=/usr/local \
           -Discsi_sbindir=/usr/local/sbin \
-          -Drulesdir=/usr/local/etc/udev/rules.d \
+          -Drulesdir=/usr/lib/udev/rules.d \
           -Dc_args="$CFLAGS -I/usr/local/include -DNO_SYSTEMD -DGLOB_ONLYDIR=0" \
           output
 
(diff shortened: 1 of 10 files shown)


storage/iscsi-tools/open-iscsi/pkg.yaml
```diff
@@ -51,7 +51,7 @@ steps:
           -Dno_systemd=true \
           -Dprefix=/usr/local \
           -Discsi_sbindir=/usr/local/sbin \
-          -Drulesdir=/usr/local/etc/udev/rules.d \
+          -Drulesdir=/usr/lib/udev/rules.d \
```
frezbo, line 54: I this this is the default setting in most packages, if so I guess we can drop the config even

smira (approved):

storage/iscsi-tools/open-iscsi/pkg.yaml
dsseng: Should I perhaps just remove this config line? And the similar path in nut-client?

frezbo: yes, let's do that as long as it installs the default rules at the right location

frezbo: we also need to bump extensions-validator

dsseng: open-iscsi defaults to `RULESDIR ?= $(etcdir)/udev/rules.d`, so for it I must keep the line

dsseng: And well, NUT also seems to default to etc

frezbo (approved):

frezbo (approved):

PR #505, opened 2024-10-25
Adding multipath-tools extension

aaroncirillo-cision:
This PR adds a new extension for multipath-tools. This is needed in the environment I work in, and probably others as well.

I intend for this PR to be an open conversation about this feature. I tried to implement this in a way that aligns with Talos goals and philosophy, but if there is any feedback on this please let me know.


aaroncirillo-cision:
Where can I find the requirements for properly formatting my commit messages?

multipath-tools/src/go/start_multipath.go
```diff
@@ -0,0 +1,28 @@
...
+    "fmt"
+    "os"
+    "os/exec"
+    "syscall"
+)
+
+func main() {
+    fmt.Println("Starting multipath daemon...")
+
+    //setup a chroot environment for the extension
```
frezbo, line 13: i don't think we need this, with `mountPropogation` shared should work as a normal extension service

multipath-tools/src/rootfs/etc/multipath.conf
```diff
@@ -0,0 +1,1132 @@
+defaults {
```
frezbo, line 1: this should be provided as an `ExtensionServiceConfig`

multipath-tools/src/rootfs/lib/ld-musl-x86_64.so.1
frezbo, line 1: this should just compile using `base` and put all libs under `/rootfs/usr/local/lib/containers/multipath-tools/`

aaroncirillo-cision: ok, sounds good, thanks

aaroncirillo-cision: @lusrmgr 
Not quite yet. I am running this extension internally right now and it is working for us. The extension I built was based on me grabbing binaries from Alpine and sticking them into an extension image. The Talos project can't accept that. They need a build from source integrated into their build system. So I do want to go back and clean up what I have and get a build running in their build system.

I've seen a few github threads on this topic actually. It looks like many of the CSI drivers for fiber channel arrays will have this same requirement. So getting this done would benefit everyone.

You may or may not have interest in the thread here as well discussing the same topic: https://github.com/hpe-storage/csi-driver/issues/379

frezbo: there's also work to actually have `/etc/iscsi/initiatorname.iscsi` here: https://github.com/siderolabs/talos/pull/10162 and some rework of iscsi-tools here: https://github.com/siderolabs/extensions/pull/577

bzub: FWIW I've started working on this too. So far I've got multipath-tools compiling. The work in this PR is a huge help. I can submit a PR soon as well.

aaroncirillo-cision: Your branch looks great, much better than what I did. I'm glad someone who had a better understanding of the build system took a look here. This was really my first encounter with Talos extensions and I was missing a lot of detail about how the bldr tool worked and how you could pull in dependencies from the core packages.

oliverbaehler: @bzub @aaroncirillo-cision Anything we can do here?

oliverbaehler: Created thread in Slack, for anyone interested:

https://taloscommunity.slack.com/archives/CG25RPZNE/p1740658131754019

frezbo:
> @frezbo What kind of result are you expecting from such a test/is there examples eg how you are testing the iscsi extension?

https://github.com/siderolabs/talos/blob/70612c1f9fc9056e8a3669ff10a385c4e8e03350/internal/integration/k8s/longhorn.go#L128 we have a test for iscsi using longhorn

Untersander:
For anyone interested we've got a working build of a multipath extension which works in connection with the NetApp Trident CSI with ISCSI volumes.
There is some cleanup needed and we'll probably add an additional extension and/or trident PR just for supporting binaries in that are needed by Trident.
We'll also try to create tests for it so it may be merged here.
Currently our fork only works for Trident, because we hardcode the multipath.conf, we'll try to get it into a ExtensionServiceConfig, but this probably also needs changes to Trident.

See: https://github.com/INSRapperswil/extensions/tree/slim-multipath

frezbo:
multipathd extension is now available

PR #538, opened 2024-12-03
feat: add dvb-cx23885 extension

samip5:


frezbo:
Would probably make sense to ship both firmware and drivers together

samip5:
@frezbo That makes sense, and added it too now.

drivers/dvb-cx23885/manifest.yaml
```diff
@@ -0,0 +1,11 @@
...
+metadata:
+  name: dvb-cx23885
+  version: "$VERSION"
+  author: Skyler Mäntysaari
+  description: |
+    This system extension provides the dvb kernel modules required for Hauppage WinTV-quadHD PCIe tuner built against a specific Talos version.
+    Includes the firmware required.
+  compatibility:
+    talos:
+      version: ">= v1.8.0"
```
frezbo, line 11: 1.9.0

PR #542, opened 2024-12-07
feat: add cloudflared system extension

maxnrb:
This PR adds a new system extension cloudflared

Cloudflare Tunnel securely connects resources to Cloudflare without a public IP. A lightweight daemon (cloudflared) creates outbound-only connections to Cloudflare, allowing safe access to services like HTTP, SSH, remote desktops, and other protocols.

More info: https://github.com/cloudflare/cloudflared/


network/cloudflared/manifest.yaml
```diff
@@ -0,0 +1,13 @@
...
+  version: "$VERSION"
+  author: Maxime Narbaud
+  description: |
+    Cloudflare Tunnel securely connects resources to Cloudflare without a public IP.
+    A lightweight daemon (cloudflared) creates outbound-only connections to Cloudflare,
+    allowing safe access to services like HTTP, SSH, remote desktops, and other protocols.
+    More info: https://github.com/cloudflare/cloudflared/
+  compatibility:
+    talos:
+      version: ">= v1.5.0"
```
frezbo, line 13: 1.9.0

smira: I think from pure API perspective, 1.5.0 should be fine as well? (if someone builds manually?)

frezbo: yeh, will keep as is then

frezbo (approved):

maxnrb:
Hi @frezbo, 

I've just seen your commit, and instead of retrieving the cloudflared binaries directly, you locate the sources and build the binary. 

However, cloudflared is built with a version of go ‘modified’ by Cloudflare (https://github.com/cloudflare/go), see the Dockerfile: https://github.com/cloudflare/cloudflared/blob/master/Dockerfile
So I'm afraid that some of the tunnel's functionality won't work (particularly security functionality, as their version patches the TLS stack).

I can push a new PR by building the binary from source with Cloudflare's go executable. What do you think about?

frezbo:
> Hi @frezbo,
> 
> I've just seen your commit, and instead of retrieving the cloudflared binaries directly, you locate the sources and build the binary.

We usually do that, so we build all binaries with the go version that ships with our toolchain, we pick a different one only for special cases

> 
> However, cloudflared is built with a version of go ‘modified’ by Cloudflare (https://github.com/cloudflare/go), see the Dockerfile: https://github.com/cloudflare/cloudflared/blob/master/Dockerfile So I'm afraid that some of the tunnel's functionality won't work (particularly security functionality, as their version patches the TLS stack).

Are you sure https://github.com/cloudflare/cloudflared/blob/master/Dockerfile#L4 this just seems to use upstream golang image

frezbo:
> Are you sure https://github.com/cloudflare/cloudflared/blob/master/Dockerfile#L4 this just seems to use upstream golang image

nevermind, saw the lines after that, in that case I guess we can fallback to pre-built binaries, wdyt @smira

smira:
> > Are you sure [cloudflare/cloudflared@`master`/Dockerfile#L4](https://github.com/cloudflare/cloudflared/blob/master/Dockerfile?rgh-link-date=2024-12-12T03%3A57%3A12Z#L4) this just seems to use upstream golang image
> 
> nevermind, saw the lines after that, in that case I guess we can fallback to pre-built binaries, wdyt @smira

idk, mixed feelings about that. but I guess if someone runs `cloudflared`, they already trust their security to Cloudflare, so using upstream binaries is fine.

maxnrb:
Hi, so finally ?

frezbo:
> Hi, so finally ?

yes, we would use the binaries

maxnrb:
Okay, I'll push a fix then

PR #545, opened 2024-12-10
feat: add dvb-cx23885 extension

frezbo:
This allows for using the Hauppage WinTV-quadHD PCIe tuner, and this bundles the firmware for the tuner as well.

Re-created from #538 since initial PR created from org


smira (commented):
need to update README.md?

smira (approved):

PR #552, opened 2024-12-14
fix: used cloudflared binaries instead of building from sources

maxnrb:
Hi, here are the changes linked to this PR https://github.com/siderolabs/extensions/pull/542, to put the binaries back rather than building from the sources.

By the way, I don't understand why you changed the version variable names ([in your commit of the previous PR](https://github.com/siderolabs/extensions/commit/1dd6c364b8aae54a891ef1b371417e4b6b263030)) from ‘.VERSION’ to ‘.CLOUDFALRED_VERSION’. In the end, in the manifest.yaml, you end up replacing some variables with ‘.CLOUDFLARED_VERSION’ and leaves others with ‘.VERSION’, which isn't clear or consistent.
I've replaced everything with ‘.CLOUDFLARED_VERSION’, and deleted the vars.yml file containing the ‘.VERSION’ vars, which is now useless.

Finally, could you remove the current ghcr.io/siderolabs/cloudflared:2024.12.1 image, which is therefore non-compliant, or will it be automatically replaced?

Thanks in advance :)

### DIFF
--- a/network/cloudflared/manifest.yaml
+++ b/network/cloudflared/manifest.yaml
@@ -1,7 +1,7 @@
 version: v1alpha1
 metadata:
   name: cloudflared
-  version: "$VERSION"
+  version: "$CLOUDFLARED_VERSION"
   author: Maxime Narbaud
   description: |
     Cloudflare Tunnel securely connects resources to Cloudflare without a public IP.
--- a/network/cloudflared/pkg.yaml
+++ b/network/cloudflared/pkg.yaml
@@ -5,30 +5,25 @@ dependencies:
   - stage: base
 steps:
   - sources:
-      - url: https://github.com/cloudflare/cloudflared/archive/refs/tags/{{ .CLOUDFLARED_VERSION }}.tar.gz
-        destination: cloudflared.tar.gz
-        sha256: 74794fbcdd7b71131799100d493cf70a8e126cb109f3d9e2abce55593df6a737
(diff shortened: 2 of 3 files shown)


frezbo:
> By the way, I don't understand why you changed the version variable names ([in your commit of the previous PR](https://github.com/siderolabs/extensions/commit/1dd6c364b8aae54a891ef1b371417e4b6b263030)) from ‘.VERSION’ to ‘.CLOUDFALRED_VERSION’. In the end, in the manifest.yaml, you end up replacing some variables with ‘.CLOUDFLARED_VERSION’ and leaves others with ‘.VERSION’, which isn't clear or consistent.
I've replaced everything with ‘.CLOUDFLARED_VERSION’, and deleted the vars.yml file containing the ‘.VERSION’ vars, which is now useless.

That's just some internal workings, but we need the `vars.yaml`, I'll fix that up

> Finally, could you remove the current ghcr.io/siderolabs/cloudflared:2024.12.1 image, which is therefore non-compliant, or will it be automatically replaced?

Each talos release has a matching extension matched by digest, so the old one would not be picked at all, anyways we can clean it up

maxnrb:
Okay, thanks for the informations

frezbo (approved):

frezbo:
> Okay, thanks for the informations

Actually your fixes are good, it was just following existing pkgs syntax, all good

PR #574, opened 2024-12-28
fix: dvb cx23885 module loading

samip5:
I believe that because this is not included, it's not able to load it properly.


samip5:
It does seem to exist.

PR #577, opened 2025-01-05
feat: rework iscsi-tools

frezbo:
* `tgtd` is only needed if exposing `iSCSI` targets from the host, not needed when using as a client.
* Drop `isns` since we never ran `isns` daemon
* Drop `iscsid-wrapper` since Talos now manages `iqn`, ref: https://github.com/siderolabs/talos/pull/10162


storage/iscsi-tools/iscsid-wrapper/main.go
```diff
@@ -23,18 +23,19 @@ func main() {
 
 	cmd.Stdout = &cmdOut
 
-	if _, err := os.Stat("/etc/iscsi/initiatorname.iscsi"); err != nil {
```
smira, line 26: how would this work on upgrades? if there already was a name provided?

runningman84:
Please test this with longhorn...

frezbo:
> Please test this with longhorn...

already here https://github.com/siderolabs/talos/pull/10087

voxxit:
And Synology CSI, which frankly needs docs revisions. I have a Synology NAS I can use to help test, if necessary.

storage/iscsi-tools/iscsid-wrapper/main.go
frezbo: yeh, this would be a change, I've added a release note on Talos side

storage/iscsi-tools/vars.yaml
```diff
@@ -1,4 +1,4 @@
-VERSION: v0.1.6
+VERSION: v0.2.0
 # renovate: datasource=github-tags depName=open-iscsi/open-iscsi
 OPEN_ISCSI_VERSION: 2.1.10
```
smira, line 3: OPEN_ISNS_VERSION can now be dropped?

storage/iscsi-tools/iscsid-wrapper/main.go
smira: how did it write this file before? where did it go (given that `/etc` is ro)?

storage/iscsi-tools/vars.yaml
frezbo: ah yes

storage/iscsi-tools/iscsid-wrapper/main.go
frezbo: it was a mount from `/system` but it never worked since `'system` is tmpfs

smira: ah, so it was re-created each time...

smira (approved):

PR #579, opened 2025-01-10
fix: add both Aenix and Enix in MAINTAINERS file

donch:


smira:
@kvaps as you're the original maintainer?

kvaps (approved):
Yes, I'm working for Aenix, so it's better to display company name instead of my personal name.

I can approve this PR.

smira (approved):

Issue #578, opened 2025-01-08
How to pass host devices into kata-containers raw, without virtio-fs?


gerhard:
This branches off:
- https://github.com/dagger/dagger/issues/9319

All config - including the Talos v1.9.1 cluster managed by Omni - is captured in that issue.

I am opening this issue here instead of https://github.com/kata-containers/kata-containers since this is specific to the kata-containers extension as configured in Talos (i.e. containerd v2 runtime handler).

I have a bunch of host devices which I want to pass into various kata containers. I **think** that I need to change the `privileged_without_host_devices = false` setting in some way, but I am not sure.

My starting point is a loopback host device - `/dev/loop4: [3145894]:2 (/dev/tmpfs/10G/data.img)` - which I need to pass into a container as is, no virtio-fs (it clashes with overlayfs in my case).

How could I do this @fidencio @smira @frezbo ?

---
FWIW, I am working towards having a few more devices like this one, but backed by different storage implementations (local NVMe, Regatta Storage, etc.). Ideally, these would be host mount points that I just pass through - e.g. `/dev/loop4 on /var/lib/dagger/10G.ramdisk type ext4 (rw,noatime,nodiratime,nobarrier)` - but I **think** that solving that problem is more involved. WDYT?

PR #580, opened 2025-01-10
fix: unable to override runtime defaults

frezbo:
This fixes the issue when `.machine.files` using
`/etc/cri/conf.d/20-customization.part` is sometimes unable to merge the values.

The fix is to suffix the the files with `10-` prefix so the customization with `20-` takes precendence when merging.

`gvisor-debug` is prefixed with `11-` since it should overwrite the gvisor defaults. This also fixes where the order of gvisor extensions mattered previously.

Fixes: #578


smira (approved):

gerhard:
Thank you!

Looking forward to testing this in Talos 1.9.x via Omni as soon as it becomes available.

Issue #562, opened 2024-12-20
nvme-cli


frezbo:
nvme-cli extension

frezbo:
Things needed on talos side:

* `configfs` - `mount -t configfs configfs /sys/kernel/config`
* way to provide `/usr/local/etc/nvme/discovery.conf`
* handling of systemd units

rothgar:
What do the systemd units do?

jabenedicic:
Been looking at this for Trident CSI, happy to help if I can at all.

frezbo:
> What do the systemd units do?

https://github.com/linux-nvme/nvme-cli/tree/master/nvmf-autoconnect/systemd from here:

We could handle most expect for https://github.com/linux-nvme/nvme-cli/blob/master/nvmf-autoconnect/systemd/nvmf-connect%40.service.in since it  gets called by udev here: https://github.com/linux-nvme/nvme-cli/blob/master/nvmf-autoconnect/udev-rules/70-nvmf-autoconnect.rules.in#L18 which is common with many udev rules

frezbo:
> Been looking at this for Trident CSI, happy to help if I can at all.

if you could point to some docs on installation and using with nvme we could add that as a test

PR #587, opened 2025-01-20
feat: add nvme cli

frezbo:
Add nvme cli.

Fixes: #562


smira (approved):
README should be updated as well! :)

PR #596, opened 2025-02-05
feat: add zerotier extension to Talos

ht-danielgo:
- This builds a Zerotier 1.14.2 (latest as of writing) for use as a Talos extension
- Tested OK on a Talos 1.8.4 host
- Requires a service config that looks like this
```
apiVersion: v1alpha1
kind: ExtensionServiceConfig
name: zerotier
configFiles:
  - content: ' '
    mountPath: /var/lib/zerotier-one/networks.d/<network id>.conf
  - content: <zerotier-idtool secret file content>
    mountPath: /var/lib/zerotier-one/identity.secret
  - content: <zerotier-idtool public file content>
  - mountPath: /var/lib/zerotier-one/identity.public
```


network/zerotier/zerotier.yaml
```diff
@@ -0,0 +1,55 @@
...
+      - bind
+      - ro
+  ## Zerotier needs to write to this to create the interfaces
+  - source: /dev/net/tun
+    destination: /dev/net/tun
+    type: bind
+    options:
+      - bind
+      - rw
+  ## Zerotier state.
```
frezbo, line 47: are these the same files being passed as ExtensionServiceConfig?

network/zerotier/zerotier.yaml
```diff
@@ -0,0 +1,55 @@
...
+  - source: /var/lib/zerotier-one/
+    destination: /var/lib/zerotier-one/
+    type: bind
+    options:
+      - bind
+      - rw
+restart: always
+logToConsole: true
```
frezbo, line 55: not sure if we need this, this is only needed for mostly debugging

ht-danielgo: this is true, not needed going forward.

network/zerotier/zerotier.yaml
ht-danielgo: Yeh so zerotier is not great - basically `zerotier-one` optionally takes a <path> to use as its homedir. From there you need to supply the 3 files I mentioned in the ExtensionServiceConfig. By default it likes /var/lib/zerotier-one and my brief tests I had trouble using a more container local path. 

If you dont supply files like demonstrated, you need to run the zerotier-cli executable to configure it (which ultimately just writes these same files)

ht-danielgo: I'll check, pretty sure it writes further state files in here.

ht-danielgo: Yep, tracks a lot of zerotier state
```
talosctl list  /var/lib/zerotier-one -r --talosconfig=./talosconfig --nodes 167.x.x.x
NODE            NAME
167.x.x.x .
167.x.x.x   authtoken.secret
167.x.x.x   controller.d
167.x.x.x   controller.d/network
167.x.x.x  controller.d/trace
167.x.x.x   identity.public
167.x.x.x   identity.secret
167.x.x.x  metrics.prom
167.x.x.x   metricstoken.secret
167.x.x.x   networks.d
167.x.x.x   networks.d/deadbeef.conf
167.x.x.x  peers.d
167.x.x.x   peers.d/deadbeef.peer
167.x.x.x   peers.d/deadbeef.peer
167.x.x.x   planet
167.x.x.x   zerotier-one.pid
167.x.x.x   zerotier-one.port
```

network/zerotier/zerotier.yaml
```diff
@@ -0,0 +1,54 @@
...
+      - bind
+      - ro
+  # more libs
+  - source: /usr/lib
+    destination: /usr/lib
+    type: bind
+    options:
+      - bind
+      - ro
+  ## Required for zerotier. Ip addr and other commands
```
frezbo, line 33: talos doesn't ship any of those commands so I'm not sure sbin mount has any effect

ht-danielgo: This is fair, i took it straight from the WG extension, but will remove

ht-danielgo:
Just want to pause on this for a bit as I have a better way to deal with the configuration of the ZT credentials that will make configuration much easier, should have it ready end of week

rob-htl:
This should be ready to go now, changes:

* Add zerotier-wrapper as extension entrypoint, this wrapper:
  * handles PID cleanup and creation
  * works instead of the upstream zerotier symlinks for zerotier-cli and zerotier-idtool
  * takes configuration from ENV vars ZEROTIER_NETWORK and optionally ZEROTIER_IDENTITY_SECRET
  * ZEROTIER_NETWORK must be set, this is the ID of the network that zerotier attempts to join after start up (typically a manual process)
  * If ZEROTIER_IDENTITY_SECRET is optionally set this is written out and used by zerotier to authenticate as the node
  * If ZEROTIER_IDENTITY_SECRET is not set a new identity is created
  * logs the various lifecycle steps in a verbose way

* Remove unused mounts (zerotier is compiled statically so /lib etc aren't needed)
* Removes the aforementioned symlinks for zerotier-cli and zerotier-idtool as they aren't relevant without shell
* Adds some usage docs in network/zerotier/README.md
* Adds the renovate comment for zerotier version
* Rebased to main and resolved conflicts

network/zerotier/pkg.yaml
```diff
@@ -0,0 +1,53 @@
+name: zerotier
+variant: alpine
```
frezbo, line 2: any reason  alpine is used, should just build fine with our toolchain

network/zerotier/zerotier-wrapper/main.go
```diff
@@ -0,0 +1,232 @@
...
+			if err := joinNetwork(network); err != nil {
+				log.Printf("failed to join network: %v", err)
+			} else {
+				log.Printf("joined network %s", network)
+			}
+		}()
+	}
+
+	// Start zerotier-one process.
+	cmd := exec.Command(zerotierBinPath, "-U", zerotierPath)
```
frezbo, line 60: this is making this more complicated, this can be replaced via `unix.Exec` making all the pid handling not needed

network/zerotier/zerotier-wrapper/main.go
```diff
@@ -0,0 +1,232 @@
+package main
```
frezbo, line 1: so this basically makes generating the pub key easier?

frezbo: Just trying to see if needing the wrapper makes sense if user could provide secret and pub key as extension service config

network/zerotier/zerotier.yaml
```diff
@@ -0,0 +1,36 @@
...
+    - etcfiles
+  - configuration: true
+container:
+  entrypoint: /usr/local/bin/zerotier-wrapper
+  security:
+    writeableRootfs: false
+    writeableSysfs: true
+  mounts:
+  # Zerotier binaries
+  - source: /usr/local/lib/containers/zerotier/usr/local/bin
```
frezbo, line 16: let's not do this and just keep binaries in `/usr/local/lib/containers/zerotier/usr/local/bin`

network/zerotier/pkg.yaml
rob-htl: I've updated to use the toolchain, it didn't build for me using the toolchain originally hence alpine but it builds just fine now - so that's great

network/zerotier/zerotier-wrapper/main.go
rob-htl: `exec.Command` was due to using `zerotier -q join <network_id>` requiring the service to be up. However all "joining" the network entails is creating an empty config file with the network name, so I have done that to join instead and removed `exec.Command` and all PID handling.

network/zerotier/zerotier-wrapper/main.go
rob-htl: In short yes. There has to be a unique identity per node.

The [usage guide from Zerotier](https://docs.zerotier.com/start#setup-the-zerotier-app) boils down to:

* Install from Zerotier repos
* (the debian package generates the identity post-install - presumably the others do too as it's not handled by the following commands)
* use `zerotier-cli join <network_id>` (i.e `zerotier-one -q join <network_id>`)
* authorize the node at the zerotier end (the most common way to do this is via the web interface but there are other ways to do it programatically)
* the node is connected

The alternative approach (without the wrapper) is for an admin to have zerotier installed on their workstation and each time a new node is added to a Talos cluster generate the identities one by one (`zerotier-idtool` / `zerotier-one -i`) and create a MachineConfig per node.   On the Talos node there will be no error, the zerotier service will start and display this output only which could cause user confusion.
```
Starting Control Plane...
Starting V6 Control Plane...
```

This wrapper allows the admin to have one config patch per cluster for the zerotier network id (with a feature to allow the identity to be provided via env per node and be validated), and it provides more robust logging output.

network/zerotier/zerotier.yaml
rob-htl: updated

network/zerotier/zerotier-wrapper/main.go
frezbo: we can also split this into two services, on that runs once only, and it doesn't run if file already exists

rob-htl: I don't quite follow, are you suggesting something like the following?

* `zerotier-setup`: create identity, join network etc etc
* `zerotier-wrapper`: checks that setup has been done and `unix.Exec` `zerotier-one`


e.g. 

```
container:
  entrypoint: /bin/sh
  args:
    - -c
    - |
      /usr/local/bin/zerotier-setup && /usr/local/bin/zerotier-wrapper
```

rob-htl: > we can also split this into two services, on that runs once only, and it doesn't run if file already exists

@frezbo are you able to clarify what you want here please? I'm keen to get it done and resolved

frezbo: sorry, I'll try to get back to this by this week or next, don't worry, this will go in 1.10, just finishing other tasks

frezbo (approved):

frezbo:
@rob-htl would you mind signing off the commit, you can pull the changes i pushed and do a `git commit -s --amend --no-edit` and `git push --force`

rob-htl:
> @rob-htl would you mind signing off the commit, you can pull the changes i pushed and do a `git commit -s --amend --no-edit` and `git push --force`

Done - thanks - I also updated `MAINTAINERS.md` as I've taken this over from [ht-danielgo](https://github.com/ht-danielgo)

PR #604, opened 2025-02-11
feat: update talos-vmtoolsd to v1.0.0

smira:
It has just been released.

### DIFF
--- a/guest-agents/vars.yaml
+++ b/guest-agents/vars.yaml
@@ -7,6 +7,6 @@ PCRE2_VERSION: 10.42
 # renovate: datasource=git-tags depName=https://gitlab.com/xen-project/xen-guest-agent.git
 XEN_GUEST_AGENT_VERSION: 5c274e651c29f92fc0c418fda486373b0f34f0da
 # renovate: datasource=github-releases depName=siderolabs/talos-vmtoolsd
-TALOS_VMTOOLSD_VERSION: v0.6.1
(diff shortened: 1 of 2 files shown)


smira:
Forked from #603.

dsseng (approved):

smira:
The problem here is that this extension is amd64 only, and the image is now properly tagged for that, so the build fails.

smira:
> The problem here is that this extension is amd64 only, and the image is now properly tagged for that, so the build fails.

Solved that, now arm64 extension would be empty/non-functional as it should be.

jonkerj:
Yeah, I'm a bit devided on darwin and arm64:
- in practise, this will only be run on VMWare-hosted talos, which is typically linux/amd64
- but in theory, ESX on ARM does exist
- You could run `talos-vmtoolsd talosquery --talos-config <x> --talos-node <y>` on any machine to check stuff

I you think this warrants additional targets (eating artifact space, using build time and creating false positives during testing), I'm happy to add those.

smira:
I don't have much idea, but last time I checked VMWare Go libraries don't build on arm64, so Talos VMWare platform is no-op on arm64.

If you want extension to work, you need to make sure it builds for arm64 first, and then we can remove this hack.

jonkerj:
Let's wait for actual vmware/arm64 users to raise issues about this then :-)

PR #605, opened 2025-02-11
feat: add Nebula extension to Talos

iamwacko:
Nebula is an overlay network system, sort of like Tailscale or OpenZiti. The extension builds, and I made an iso with it that I'll test when I have time.

### DIFF
--- a/network/nebula/README.md
+++ b/network/nebula/README.md
@@ -1,0 +1,49 @@
+# Nebula
+
+https://github.com/slackhq/nebula
+
+## Installation
+
+See [Installing Extensions](https://github.com/siderolabs/extensions#installing-extensions).
+
+## Usage
+
+Configure the extension via `ExtensionServiceConfig` document. You can add any nebula related configuration and these will be executed at Nebula startup.
+
+```yaml
+---
+apiVersion: v1alpha1
+kind: ExtensionServiceConfig
+name: nebula
+configFiles:
+  - content: |
+        pki:
+        ca: /usr/local/etc/nebula/ca.crt
+        cert: /usr/local/etc/nebula/node.crt
+        key: /usr/local//etc/nebula/node.key
+    mountPath: /usr/local/etc/nebula/config.yml
+  - content: |
+        -----BEGIN NEBULA CERTIFICATE-----
+        -----END NEBULA CERTIFICATE-----
+    mountPath: /usr/local/etc/nebula/ca.crt
+   - content: |
+        -----BEGIN NEBULA CERTIFICATE-----
+        -----END NEBULA CERTIFICATE-----
+     mountPath: /usr/local/etc/nebula/node.crt
+   - content: |
+     mountPath: /usr/local/etc/nebula/node.key
+
+```
+
+Then apply the patch to your node's MachineConfigs
+```bash
+talosctl patch mc -p @nebula-config.yaml
+```
+
+You will then be able to verify that it is in place with the following command
+```bash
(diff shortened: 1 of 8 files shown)


iamwacko:
It seems to work. It's already works in containers (https://github.com/slackhq/nebula/tree/master/docker), so it's unsurprising.

iamwacko:
There's no Contributors.md, do I have to make an issue to go with the PR or something?

network/nebula/README.md
```diff
@@ -0,0 +1,49 @@
...
+Configure the extension via `ExtensionServiceConfig` document. You can add any nebula related configuration and these will be executed at Nebula startup.
+
+```yaml
+---
+apiVersion: v1alpha1
+kind: ExtensionServiceConfig
+name: nebula
+configFiles:
+  - content: |
+        pki:
```
frezbo, line 20: is `pki` an empy key or whether `ca`, `cert` and `key` should be under `pki`

network/nebula/nebula.yaml
```diff
@@ -0,0 +1,47 @@
...
+container:
+  entrypoint: /usr/local/bin/nebula
+  args: 
+    - -config 
+    - /usr/local/etc/nebula/config.yml 
+  environment:
+  - PATH=/sbin:/usr/local/bin
+  security:
+    writeableRootfs: false
+    writeableSysfs: true
```
frezbo, line 18: is this needed?

network/nebula/nebula.yaml
```diff
@@ -0,0 +1,47 @@
...
+  mounts:
+  # libs
+  - source: /lib
```
frezbo, line 21: if nebula is compiled as a static binary I guess these `lib` and `sbin` mounts are not needed

network/nebula/pkg.yaml
```diff
@@ -0,0 +1,50 @@
...
+  - network: none
+    build:
+      - |
+        cd nebula
+        make bin
+    install:
+      - |
+        mkdir -p /rootfs/usr/local/lib/containers/nebula/usr/local/bin/
+        cp -pr nebula/nebula /rootfs/usr/local/lib/containers/nebula/usr/local/bin
+        cp -pr nebula/nebula-cert /rootfs/usr/local/lib/containers/nebula/usr/local/bin
```
frezbo, line 36: is `nebula-cert` binary used for the service?

network/nebula/README.md
iamwacko: `ca`, `cert`, and `key` are supposed to be under `pki`

network/nebula/pkg.yaml
iamwacko: I'd assumed it was, but looking at the documentation it seems like it's not needed

network/nebula/nebula.yaml
iamwacko: I'm pretty sure it needs Sysfs, since it uses tun. It shouldn't need to write to root.

If you mean the PATH, I just took that from Tailscale, I could probably remove it.

network/nebula/nebula.yaml
iamwacko: Okie dokie, I'll remove the mounts

network/nebula/README.md
frezbo: okay, I'll fix the example, here pki and ca are all under same indentation

frezbo (approved):

PR #608, opened 2025-02-13
fix: update nvidia-container-cli and make it build with current Go

dsseng:
Go 1.24 update in Wolfi made this not build due to deprecation of older language features apparently.

Before the fix is upstreamed keep it in-tree, remove when it lands in a tagged release.

Signed-off-by: Dmitry Sharshakov <dmitry.sharshakov@siderolabs.com>


smira (approved):

nvidia-gpu/vars.yaml
```diff
@@ -7,8 +7,10 @@ NVIDIA_DRIVER_PRODUCTION_VERSION: 550.144.03
 CONTAINER_TOOLKIT_VERSION: v1.17.3
```
frezbo, line 7: we need to bump this too, both always gets bumped in unison, and I think nvidia won't work when there's a version mismatch

dsseng: Whoops, right, thanks, fixing

PR #638, opened 2025-03-07
feat: add extenstion for panfrost module

adamcirillo:
This extension loads the panfrost module for arm mali midgard & birfrost gpus used by various sbcs. I've compiled the extension the _out folder only had the manifest.yaml and the bldr file this seemed the same as the amd gpu drm module so i'm thinking this is correct. Also is there anything else I need to do to add this to image factory or this it pick this up automatically once merged? Also updated the makefile with the kernel build that has the panfrost module.

### DIFF
--- a/drm/panfrost/pkg.yaml
+++ b/drm/panfrost/pkg.yaml
@@ -1,0 +1,49 @@
+name: panfrost
+variant: scratch
+shell: /bin/bash
+# {{ if eq .ARCH "aarch64" }} This in fact is YAML comment, but Go templating instruction is evaluated by bldr
+dependencies:
+  - stage: base
+  # The pkgs version for a particular release of Talos as defined in
+  # https://github.com/siderolabs/talos/blob/<talos version>/pkg/machinery/gendata/data/pkgs
+  - image: "{{ .BUILD_ARG_PKGS_PREFIX }}/kernel:{{ .BUILD_ARG_PKGS }}"
+steps:
+  - prepare:
+      - |
+        sed -i 's#$VERSION#{{ .VERSION }}#' /pkg/manifest.yaml
+
+        mkdir -p /rootfs
+  - install:
+      - |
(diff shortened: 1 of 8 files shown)


drm/panfrost/pkg.yaml
```diff
@@ -0,0 +1,39 @@
...
+  # https://github.com/siderolabs/talos/blob/<talos version>/pkg/machinery/gendata/data/pkgs
+  - image: "{{ .BUILD_ARG_PKGS_PREFIX }}/kernel:{{ .BUILD_ARG_PKGS }}"
+  - image: "{{ .BUILD_ARG_PKGS_PREFIX }}/linux-firmware:{{ .BUILD_ARG_PKGS }}"
+steps:
+  - prepare:
+      - |
+        sed -i 's#$VERSION#{{ .VERSION }}#' /pkg/manifest.yaml
+
+        mkdir -p /rootfs
+  # {{ if eq .ARCH "x86_64" }} This in fact is YAML comment, but Go templating instruction is evaluated by bldr
```
frezbo, line 16: is this correct, isn't this only installing for amd64, whereas Panfrast is ARM64 ? :thinking:

adamcirillo: yes this should be for arm64 is the correct syntax 
```
# {{ if eq .ARCH "x86_64" }} This in fact is YAML comment, but Go templating instruction is evaluated by bldr
```

frezbo: we also need to push an empty manifest for amd64 so factory works, see the example here: https://github.com/siderolabs/extensions/blob/main/guest-agents/vmtoolsd-guest-agent/pkg.yaml

adamcirillo: I've made your suggested changes and removed the linux firmware as there is no binary firmware for panfrost driver, i've compiled and see the modules in the _out folder now and the stub manifest with amd64 build as well.

adamcirillo: is everything all good now?

frezbo (approved):

PR #640, opened 2025-03-08
chore: fix Nebula image link in README

josephglanville:
Trivial fix to README to fix image link for Nebula extension.


smira (approved):
thank you!

PR #641, opened 2025-03-09
feat: add helper code for supporting the RevolutionPi4 as revpi-firmware

c0ffee:
Using factory assigned MAC address from EEPROM instead of compute module address

this depends on https://github.com/siderolabs/sbc-raspberrypi/pull/44


firmware/revpi-firmware/src/main.go
```diff
@@ -0,0 +1,76 @@
...
+		os.Exit(1)
+	}
+
+	file, err := os.Open("/sys/firmware/devicetree/base/hat/custom_5")
+	if err != nil {
+		log.Fatalf("Failed to open file: %v", err)
+	}
+	defer file.Close()
+
+	scanner := bufio.NewScanner(file)
```
frezbo, line 40: can we just do `os.Readfile`?

c0ffee: sure, makes sense :) thanks for the feedback

frezbo (approved):

PR #647, opened 2025-03-14
feat: add ctr extension

### LINKED ISSUE #622: Create extension with `ctr` binary
We currently [remove `ctr`](https://github.com/siderolabs/pkgs/blob/7774b08f03f5c096efdcc7863260916d78a7b8a9/containerd/pkg.yaml#L31) from the default containerd build so users won't rely on the binary for managing containers on the host.

Some workloads require the binary for various host level scanning. We should probably keep the binary off of the base OS, but we should also enable users who need it to use it.

If we create an extension it will be built and tracked with talos builds and should remain compatible. Are there any other common container runtime tools needed for 3rd party software we should consider adding to an optional extension?

tomaszjonak:
This PR provides containerd ctr binary in a form of a stanadlone talos extension. Counterpart to https://github.com/siderolabs/pkgs/pull/1186.

cc @redhatrises.

Closes #622


smira:
Same questions as in https://github.com/siderolabs/pkgs/pull/1186, and also pkgs path is better than downloading binary release.

redhatrises:
@smira with [siderolabs/pkgs#1186](https://github.com/siderolabs/pkgs/pull/1186) merged. Do we need to make changes to this PR?

smira:
@redhatrises I don't know, if you want `ctr` as a system extension, you need to make a PR to extensions - either new one, or updating this one? I don't know if you and @tomaszjonak are working together?

tomaszjonak:
@smira yes, we are working together. I will update this PR as needed.

> and also pkgs path is better than downloading binary release 

I'm not versed enough in talos build system. Could you expand on how this should be done. I.e is there another extension that utilizes such flow.

smira:
I don't know something that matches 100%, but e.g. https://raw.githubusercontent.com/smira/extensions/e974d6e1fcd28b064c37c4c2bec6c3cc8159f6c2/firmware/qlogic-firmware/pkg.yaml

So you take `containerd` container image built by `pkgs` and just extract `ctr` from it, and put it under `/rootfs/usr/bin`. Should be relatively simple.

You also might need to bump PKGS to be the latest (use `git describe --tags` in `pkgs` repo) in `.kres.yaml` and `make rekres`.

redhatrises:
@smira after talking to @tomaszjonak, this PR is ready for review.

smira (approved):
thank you!

PR #660, opened 2025-03-31
feat: add an extension for Amazon ENA network driver

smira:
Fixes https://github.com/siderolabs/talos/issues/10246


drivers/amazon-ena/vars.yaml
```diff
@@ -0,0 +1 @@
+VERSION: "{{ .ENA_VERSION }}-{{ .BUILD_ARG_TAG }}"
```
frezbo, line 1: i guess we can start dropping the build arg tag?

frezbo (approved):

PR #663, opened 2025-04-03
fix: gvisor shutdown

smira:
This is an attempt to fix gvisor by updating the patch from https://github.com/monogon-dev/monogon/blob/main/third_party/com_gvisor_dev_gvisor/gvisor-containerd-compat.patch.


smira:
The outcome is that runsc is still hanging

smira:
We don't need this

PR #675, opened 2025-04-25
feat(bcache): create bcache extension

mtb-xt:
Hi team, I'm proposing to add the bcache kernel module as an extension.

Talos already has a DRBD extension, which allows us to use Linstor storage.
Linstor also supports using **bcache** as a caching layer [(example config)](https://linbit.com/blog/tutorial-stacked-block-storage-in-linbit-sds-aka-linstor/).

This is a very simple extension, which adds a kernel module (I used binfmt_misc as an example).

Please let me know if any more changes are required.

### DIFF
--- a/storage/bcache/files/modules.txt
+++ b/storage/bcache/files/modules.txt
@@ -1,0 +1,4 @@
+modules.order
+modules.builtin
+modules.builtin.modinfo
+kernel/drivers/md/bcache/bcache.ko
--- a/Makefile
+++ b/Makefile
@@ -50,7 +50,7 @@ COMMON_ARGS += --build-arg=TOOLS_PREFIX="$(TOOLS_PREFIX)"
 # extra variables
 
 EXTENSIONS_IMAGE_REF ?= $(REGISTRY_AND_USERNAME)/extensions:$(TAG)
-PKGS ?= v1.11.0-alpha.0
+PKGS ?= v1.11.0-alpha.0-2-gcb108a5
 PKGS_PREFIX ?= ghcr.io/siderolabs
 TOOLS ?= v1.10.0
 TOOLS_PREFIX ?= ghcr.io/siderolabs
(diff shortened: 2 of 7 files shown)


storage/bcache/files/modules.txt
frezbo, line 1: here we'd need two files `modules-amd64.txt` and `modules-arm64.txt` and copy them accordingly based on arch, another option is to ship it with talos itself, the uncompressed module is just 800K

mtb-xt: No other extension does that, except for usb-modem-drivers, so I didn't know that's required.
I think the module list is the same, so, I need to add two identical modules.txt files for both arches?

frezbo: it's not the same, see https://github.com/siderolabs/pkgs/pull/1213#issuecomment-2829434625

mtb-xt: Yep, I saw that, but **crc64.ko** is not a bcache dependency. It needs to be included anyway? Sorry if I misunderstand how it's used.

```
❯ modinfo bcache
filename:       /lib/modules/6.14.3-zen1-1-zen/kernel/drivers/md/bcache/bcache.ko.zst
license:        GPL
author:         Kent Overstreet <kent.overstreet@gmail.com>
description:    Bcache: a Linux block layer cache
srcversion:     DCC523B5D1CD7F3FB1190AC
depends:        
intree:         Y
```

frezbo: I'm not sure, it got enabled when changing bcache to be a module, so I would include it

frezbo: I think it's easier to add this to core talos
 small module

mtb-xt: I think I got it, because in Talos kernel, crc64 is a module in amd64, it needs to be included specifically, but on arm64 it's builtin.

mtb-xt: > I think it's easier to add this to core talos small module

of course. should I close this one and make another PR in pkgs?

frezbo: yes, I guess you meant talos? Just needs updating here: https://github.com/siderolabs/talos/blob/main/hack/modules-amd64.txt and https://github.com/siderolabs/talos/blob/main/hack/modules-arm64.txt

mtb-xt: @frezbo https://github.com/siderolabs/talos/pull/10800 - thank you for your patience with me!

PR #710, opened 2025-05-27
fix(glibc): use lib64 for better compatibility

jfroy:
This patch changes the glibc extension to use the `lib64` directory for libraries instead of `lib`. This better reflects a standard Fedora/FHS/LSB distribution and fixes issues with a few containers, with the Nvidia Linux installer, and with the Nvidia container toolkit.

The Nvidia container toolkit mounts libraries at the same path as the host system and runs `ldconfig` to update the cache. Mounting 64-bit libraries in `lib` is wrong on Fedora-like _and_ Ubuntu-like distributions.

### DIFF
--- a/misc/glibc/ld.so.conf
+++ b/misc/glibc/ld.so.conf
@@ -1,1 +1,2 @@
+/usr/local/glibc/usr/lib64
 /usr/local/glibc/usr/lib
--- a/misc/glibc/pkg.yaml
+++ b/misc/glibc/pkg.yaml
@@ -24,12 +24,16 @@ steps:
         mkdir build
         cd build
 
+        echo "libc_cv_slibdir=/usr/local/glibc/usr/lib64" > config.cache
         ../configure \
           --prefix=/usr/local/glibc \
-          --libdir=/usr/local/glibc/usr/lib \
-          --libexecdir=/usr/local/glibc/usr/lib \
-          --enable-stack-protection=strong \
-          --disable-werror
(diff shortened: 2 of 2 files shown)


frezbo (approved):

jfroy:
Yeah I had a suspicion this would require some tweaks to the existing NVIDIA extensions. I'll take a look over the next few evenings. Totally OK to revert until the tests are sorted out.

jfroy:
> > Yeah I had a suspicion this would require some tweaks to the existing NVIDIA extensions. I'll take a look over the next few evenings. Totally OK to revert until the tests are sorted out.
> 
> This is the error:
> 
> ```
>   Warning  Failed     26s   kubelet            Error: failed to create containerd task: failed to create shim task: OCI runtime create failed: runc create failed: unable to start container process: error during container init: error running prestart hook #0: exit status 127, stdout: , stderr: Auto-detected mode as 'legacy'
> /usr/local/bin/nvidia-container-cli: error while loading shared libraries: libnvidia-container.so.1: cannot open shared object file: No such file or directory
> ```
> 
> I'm going to revert this until I can spend time to go figure this out

It's probably the state of the base overlay for the toolkit extension, and the arguments passed to the installer; I am speculating from what I had to do in my experimental driver container extension:

https://github.com/siderolabs/extensions/pull/476/files#diff-b4bb4081fb1bad9d93ee2a8ed537ced81391cac68514e45b6e7c913ea932720eR29

PR #717, opened 2025-05-31
feat(nvidia-fabricmanager): support Blackwell baseboards (DGX/HGX B100/B200/B300)

npdgm:
Quoting: https://docs.nvidia.com/datacenter/tesla/pdf/fabric-manager-user-guide.pdf

> On NVIDIA DGX-B200, HGX-B200, and HGX-B100 systems and later, the FabricManager package needs an additional NVLSM dependency for proper operation.
> NVLink Subnet manager (NVLSM) originated from the InfiniBand networking and contains additional logic to program NVSwitches and NVLinks.

Not running the NVLSM on Blackwell and newer fabrics with NVLink 5.0+ will result in FabricManager failing to start with error `NV_WARN_NOTHING_TO_DO`. NVSwitches will remain uninitialized and applications will fail with the `CUDA_ERROR_SYSTEM_NOT_READY` or `cudaErrorSystemNotReady` error.  The CUDA initialization process can only begin after the GPUs complete their registration process with the NVLink fabric.
A GPU fabric registration status can be verified with the command: `nvidia-smi -q -i 0 | grep -i -A 2 Fabric`. An `In Progress` state indicates that the GPU is being registered and FabricManager is likely not running or missing the NVLSM dependency. A `Completed` state is shown when the GPU is successfully registered with the NVLink fabric.

The FabricManager package includes the script `nvidia-fabricmanager-start.sh`, which is used to selectively start FabricManager and NVLSM processes depending on the underlying platform.
A key aspect of determining whether an NVLink 5.0+ fabric is present is to look for a Limited Physical Function (LPF) port in InfiniBand devices. To differentiate LPFs, the Vital Product Data (VPD) information includes a vendor-specific field called `SMDL`, with a non-zero value defined as `SW_MNG`. The first device is then selected, and it's port GUID is extracted and passed to NVLSM and FabricManager.
So both services share a configuration key that result from a common initialization process. Additionally, they communicate with each other over a Unix socket.

This patch introduces the following changes:
 * Adds NVLSM to the nvidia-fabricmanager extension.
 * Introduces a new `nvidia-fabricmanager-wrapper` program to replicate the initialization process from `nvidia-fabricmanager-start.sh`:
   * Detects NVLink 5.0+ fabrics and extracts an NVSwitch LPF port GUID. This is done by calling libibumad directly with CGO instead of parsing the output of the `ibstat` command as in the upstream script.
   * Starts FabricManager, and NVLSM only when needed.
   * Keeps both process lifecycles synchronized and ensures the Talos container will restart if either process crashes.
 * Refactors the nvidia-fabricmanager container to be self-contained, as this service does not share files with other nvidia-gpu extentions.


npdgm:
Still a work in progress. I'm waiting to run tests on DGX-B200 and older DGX systems, and refine the Talos container mounts.
But I would like to get an early review on the wrapper if possible. Thanks!

npdgm:
@frezbo I'm building rdma-core here, so perhaps #127 could be used instead?

npdgm:
Status of testing on Blackwell:
* extension service fixed with minimal bind mounts
* nvidia-fabricmanager-wrapper works as expected for LPF detection and process management
* NVLSM wasn't functional
  * not initializing (loading ?) the GRPC plugin
  * no unix socket file created for FM
  * logs with unexpected IB errors (unsure if root cause or because plugin lib wasn't found)

Fixes pending new tests:
* NVLSM bin and lib relocated to /opt to match vendor installer. dropped patching RPATH
* wrapper program is now waiting for the socket file to be created before starting FM, plus a second. vendor stratup script just does `sleep 5`

nvidia-gpu/nvidia-fabricmanager/nvidia-fabricmanager-wrapper/fixtures/dgxb200-ibstat-mlx5_0
```diff
@@ -0,0 +1,17 @@
+CA 'mlx5_0'
```
dsseng: Are these fixtures actually used by anything? I'm not an expert, but these might be unwanted in the repo (well, generally if this is not used it's not worth keeping. Are these used for testing? Or maybe some NVIDIA build scripts

nvidia-gpu/vars.yaml
```diff
@@ -19,6 +19,18 @@ NVIDIA_FABRIC_MANAGER_PRODUCTION_ARM64_SHA256: 5acd2eed01fcbba7af10ac18ec212c9e2
 NVIDIA_FABRIC_MANAGER_PRODUCTION_ARM64_SHA512: b397da3bde037ab0038bbc0ce482fe92b35240fd5ab5f95e99d176f726ad8f8220886c1197c2e85f255320c126e34accc039e5e669b8d54f8b140994c6a290e6
 NVIDIA_FABRIC_MANAGER_PRODUCTION_AMD64_SHA256: 63473386bed80ccc72ab81aa49283c48e2c9e561b99ac48b3898b6f0dedf9eae
 NVIDIA_FABRIC_MANAGER_PRODUCTION_AMD64_SHA512: 29adee0bd3d1125b32a237010f06ba1a1eff50dd51ea3492c797a51b84e73e239b72bdf810ce0df59673b678579267e046f9a5ea064788bfba6700f6e873ee2f
+# FIXME renovate
```
dsseng: I guess there's no opportunity for Renovate here, because no official repos are there. Also this update should probably be synced to the driver updates. Do we need separate LTS and Production editions (when will they have to be different versions)?

nvidia-gpu/nvidia-fabricmanager/nvidia-fabricmanager-wrapper/pkg.yaml
```diff
@@ -0,0 +1,52 @@
...
+      - |
+        cd rdma-core
+        cmake -GNinja \
+            -DENABLE_STATIC=1 \
+            -DNO_MAN_PAGES=1 \
+            -DNO_PYVERBS=1 \
+            -DCMAKE_BUILD_TYPE=Release
+        ninja lib/statics/libibumad.a
+      - |
+        CGO_ENABLED=1 go build -o nvidia-fabricmanager-wrapper nvidia-fabricmanager-wrapper
```
dsseng, line 48: Should both these components be built in the same package? However rdma-core is not used anywhere else yet, so this might be okay for now

nvidia-gpu/nvidia-fabricmanager/nvidia-fabricmanager-wrapper/fixtures/dgxb200-ibstat-mlx5_0
dsseng: If used for tests please include the tests, that'd be nice

nvidia-gpu/nvidia-fabricmanager/nvidia-fabricmanager-wrapper/nvswitch.go
```diff
@@ -0,0 +1,107 @@
...
+	}
+
+	for _, device := range devDir {
+		vpd, err := os.ReadFile(path.Join(ibPath, device.Name(), "device/vpd"))
+		if err != nil {
+			continue
+		}
+
+		// TODO: parse it like lspci does
+		if bytes.Contains(vpd, []byte("SMDL=SW_MNG")) {
```
dsseng, line 58: Is this surely NVIDIA-specific? E.g. if user has other InfiniBand devices, won't those be accidentally selected. Disregard this comment if it's not valid, I'm not knowledgeable in IB stuff

nvidia-gpu/nvidia-fabricmanager/nvidia-fabricmanager-wrapper/nvswitch.go
```diff
@@ -0,0 +1,107 @@
...
+			one port, this function returns a value of 2. In this case, the value at index 0 will be invalid as it is
+			reserved for switches. On failure, a negative value is returned.
+		*/
+		numPort := C.umad_get_ca_portguids(C.CString(lpf), &portGUIDs[0], maxPorts)
+
+		for i := range int(numPort) {
+			var guid uint64
+
+			// convert kernel __be64 to uint64
+			buf := bytes.NewReader((*[8]byte)(unsafe.Pointer(&portGUIDs[i]))[:])
```
dsseng, line 93: This looks a bit weird, could there be a simpler way to do the same?

nvidia-gpu/nvidia-fabricmanager/production/nvidia-fabricmanager.yaml
```diff
@@ -1,10 +1,7 @@
 # https://docs.nvidia.com/datacenter/tesla/pdf/fabric-manager-user-guide.pdf
 name: nvidia-fabricmanager
 container:
-  entrypoint: /usr/local/bin/nv-fabricmanager
-  args:
-   - --config
-   - /usr/local/share/nvidia/nvswitch/fabricmanager.cfg
+  entrypoint: /usr/bin/nvidia-fabricmanager-wrapper
```
dsseng, line 4: Should this change be done for all models (e.g. A100 only needs nvfm, not nvlsm)? Has this been regression-tested against a system not requiring nvlsm?

This could perhaps be done in the integration-aws-nvidia-nonfree CI workflow perhaps, if you cannot test, so that's not a big deal I guess

nvidia-gpu/nvidia-fabricmanager/nvidia-fabricmanager-wrapper/main.go
```diff
@@ -0,0 +1,156 @@
...
+		// let's wait for the actual GRPC socket to be created by the plugin
+		log.Println("nvidia-fabricmanager-wrapper: waiting for socket creation at", smSocket)
+		err := waitForFile(runCtx, smSocket, smSocketWait)
+		if err != nil {
+			log.Printf("nvidia-fabricmanager-wrapper: error waiting for socket: %v\n", err)
+		} else {
+			log.Println("nvidia-fabricmanager-wrapper: socket found at", smSocket)
+		}
+		// for safety
+		time.Sleep(time.Second)
```
dsseng, line 145: Are you waiting for nvfm to become ready? Perhaps there's no better way (some signaling from the daemon), but if there is it would be nice to consider

dsseng (commented):
Thanks, this looks like a really complex task you accomplished. Please see some of the questions I had, maybe those would help us clarify some questions

nvidia-gpu/nvidia-fabricmanager/nvidia-fabricmanager-wrapper/fixtures/dgxb200-ibstat-mlx5_0
npdgm: These fixtures were used for myself at first. But because there is room to improve the VPD lookup with proper parsing, I figured it could be used later if the code evolves. It may never happen as the current "grep like" method works well and is actually used by Nvidia themself.
I'll remove these fixtures.

nvidia-gpu/vars.yaml
npdgm: As far as I know there has been a single public release of NVLSM so far, while many driver and fabricmanager versions came out. It's not synced to the driver version, so far a single NVLSM version worked with all fabricmanager releases. It's a fair asumption that we don't need separate editions for LTS and Prod.
I'll refactor to a single set of vars.

nvidia-gpu/nvidia-fabricmanager/nvidia-fabricmanager-wrapper/pkg.yaml
npdgm: Building `libibumad.a` adds little overhead to the build process as I'm not building the whole rdma-core distribution. And it makes the nvidia-fabricmanager extension self-sufficient at runtime.

nvidia-gpu/nvidia-fabricmanager/nvidia-fabricmanager-wrapper/nvswitch.go
npdgm: Yes, `SMDL=SW_MNG` is for sure specific to NVIDIA, and more specifically to NVLink fabrics.

nvidia-gpu/nvidia-fabricmanager/nvidia-fabricmanager-wrapper/nvswitch.go
npdgm: I'm not experienced enough in Go to answer the question about simpler ways. But I spent some time on this conversion problem and it looked to be the cleanest approach. `__be64` is big endian uint regardless of the CPU platform endianness, and the `binary` package is well suited to read from arbitrary byte orders. Nevertheless it still needs to read 8 bytes from the element of the C array, which is why `unsafe.Pointer()` is used. It looks to be a common practice when using CGO as far as I can tell.
An alternative I discarded because it would be harder to maintain would have ben to add a .c file with a custom function wrapping the lib. To convert bigendian integers to hex strings.

nvidia-gpu/nvidia-fabricmanager/nvidia-fabricmanager-wrapper/main.go
npdgm: From the upstream NVIDIA script `nvidia-fabricmanager-start.sh`, invoked by systemd when fabricmanager is installed on a supported Linux distro:
```bash
if [ "$post_nvl5" = "true" ]; then
    check_required_module
    launch_nvlsm_proc
    # SM takes few mins to start the GRPC service, hence lets wait before starting FM
    sleep 5
    launch_fm_proc
else
    echo "Detected Pre-NVL5 system"
    launch_fm_proc
fi
```
When they want NVLSM to run, they just wait for 5 seconds and hope the socket is created and the service ready.
On Talos we'll get the benefit of waiting for the minimal amount of time for the socket to be created, with clear error messages on what's happening. You are correct that neither of these wait based approaches ensure the daemon is actually ready.
I have no idea how to futher check for it's true readyness. While NVLSM has roots in the OpenSM code, the Unix socket and gRPC protocol are all proprietary and AFAIK undocumented. That's why an extra `time.Sleep(time.Second)` was added just in case it takes extra milliseconds for the socket to listen properly, and because while not being good coding it's still in line with vendor pratices for starting both daemons.

nvidia-gpu/nvidia-fabricmanager/production/nvidia-fabricmanager.yaml
npdgm: > Has this been regression-tested against a system not requiring nvlsm?

No, unfortunately I can't at this time and it might not even happen. So using the CI workflow sounds like the best option.

What was tested on Blackwell is altering the IB lookup function so that no NVSwitch configuration port is found. In this case NVLSM was skipped as expected and NVFM started like it used to be. But being a Blackwell system it failed to initialize the baseboard.

nvidia-gpu/vars.yaml
dsseng: Hmm, they might manage to have to make LTS vs production versions of NVLSM incompatible in the future. Yes, too little observation to know for sure.

nvidia-gpu/nvidia-fabricmanager/nvidia-fabricmanager-wrapper/pkg.yaml
dsseng: Ah, okay, got that, it's statically linked. Good, thanks

nvidia-gpu/nvidia-fabricmanager/nvidia-fabricmanager-wrapper/nvswitch.go
dsseng: Okay. This just looks undocumented, mind adding a docs reference or at least a comment that this is a known unique ID of an NVLink fabric? Thanks

nvidia-gpu/nvidia-fabricmanager/nvidia-fabricmanager-wrapper/nvswitch.go
dsseng: Yeah, this is fine. Thanks for elaborating

nvidia-gpu/nvidia-fabricmanager/production/nvidia-fabricmanager.yaml
dsseng: > No, unfortunately I can't at this time and it might not even happen. So using the CI workflow sounds like the best option.

Okay, that should be possible. Keeping this thread open so Noel or someone else keeps need for CI test in mind as I recall that test being occasionally disabled due to pricey GPU and often failures due to no resources available in the cloud

nvidia-gpu/nvidia-fabricmanager/nvidia-fabricmanager-wrapper/main.go
dsseng: It is in line with vendor practices, yeah. No blame on you ;)

I assume extra delay should make it reliable, so accepted, thanks

nvidia-gpu/nvidia-fabricmanager/nvidia-fabricmanager-wrapper/nvswitch.go
npdgm: There's a comment above the function block, with a link to [a PDF file from NVIDIA documenting this vendor string](https://docs.nvidia.com/datacenter/tesla/pdf/fabric-manager-user-guide.pdf) as well as a text excerpt. That's were `SMDL` and `SW_MNG` came from.
Would you like me to move this inside the function block where `bytes.Contains()` happens?

For reference, here's the upstream implementation from `nvidia-fabricmanager-start.sh` using grep commands:
```bash
# Loop through each Infiniband device directory
for dir in /sys/class/infiniband/*/device; do
    # Define the path to the VPD file
    vpd_file="$dir/vpd"

    # Check if the VPD file exists
    if [ -f "$vpd_file" ]; then
        # Search for 'SW_MNG' in the VPD file
        if grep -q "SW_MNG" "$vpd_file"; then
            echo "Detected NVL5+ system"
            # Extract the Infiniband device name using parameter expansion
            device_name="${dir%/device}"  # Removes '/device' from the end of $dir
            device_name="${device_name##*/}"  # Extracts the part after the last '/'
            echo "Checking $device_name..."
            launch_ibstat $device_name
            if [ $? -ne 0 ]; then
                remove_cruft_files
                exit 1
            fi
            capture_portguids
            remove_cruft_files
            post_nvl5=true
            # if port guid was captured, return since we are only grabbing first GUID
            if [ "$capture_flag" = "true" ]; then
                echo "Using device $device_name, port $prefix_port_guid"
                break
            fi
        fi
    fi
done
```

dsseng: Ah, no problem, that's okay as is. Thanks for clarifying

nvidia-gpu/nvidia-fabricmanager/lts/pkg.yaml
```diff
@@ -2,56 +2,93 @@ name: nvidia-fabricmanager-lts
...
+                 /rootfs/usr/local/lib/containers/nvidia-fabricmanager/opt/nvidia/nvlsm/lib \
+                 /rootfs/usr/local/lib/containers/nvidia-fabricmanager/usr/share/nvidia/nvlsm
+      # nvlsm
+      - |
+        cp sm/sbin/nvlsm /rootfs/usr/local/lib/containers/nvidia-fabricmanager/opt/nvidia/nvlsm/sbin/
 
-        cp lib/libnvfm.so.1 /rootfs/usr/local/lib/libnvfm.so.1
-        ln -s libnvfm.so.1 /rootfs/usr/local/lib/libnvfm.so
+        cp sm/lib/libgrpc_mgr.so \
+           /usr/lib/libgcc_s.so.1 \
```
frezbo: why are we copying `libgcc`? is it to contain fabric manager to itself? If copying libgcc we might need to copy other so files too

nvidia-gpu/nvidia-fabricmanager/lts/pkg.yaml
```diff
@@ -2,56 +2,93 @@ name: nvidia-fabricmanager-lts
...
+      - |
+        mkdir -p /rootfs/usr/local/etc/containers
+        cp /pkg/nvidia-fabricmanager.yaml /rootfs/usr/local/etc/containers/nvidia-fabricmanager.yaml
     test:
       - |
         mkdir -p /extensions-validator-rootfs
         cp -r /rootfs/ /extensions-validator-rootfs/rootfs
         cp /pkg/manifest.yaml /extensions-validator-rootfs/manifest.yaml
-        /extensions-validator validate --rootfs=/extensions-validator-rootfs --pkg-name="${PKG_NAME}"
+        /base-rootfs/extensions-validator validate --rootfs=/extensions-validator-rootfs --pkg-name="${PKG_NAME}"
```
frezbo: we should keep the old behavior, otherwise copying `base-rootfs` adds just unnecessary time to build steps

nvidia-gpu/nvidia-fabricmanager/lts/pkg.yaml
npdgm: NVLSM dynamically loads the file. It's not seen in the ELF header, I discovered it while enabling tracing in the loader with `LD_DEBUG=...`.
[Libgcc](https://gcc.gnu.org/onlinedocs/gccint/Libgcc.html) is small and has no dependency. That's why pulling it from Wolfi just works while NVLSM was likely built on Ubuntu. And that's why I introduced Wolfi here, because it's versionned and make for a stable build. Before I copied it from the Alpine root of the builder.

nvidia-gpu/nvidia-fabricmanager/lts/pkg.yaml
npdgm: I copied this technique from `guest-agents/xen-guest-agent`.
Are you suggesting there's a way to pull `image: cgr.dev/chainguard/wolfi-base[...]` in `dependencies:` while at the same time not nuking `/extensions-validator` from the base?

nvidia-gpu/nvidia-fabricmanager/lts/pkg.yaml
npdgm: Also unless I'm mistaken, `libgcc_s.so` is nowhere to be found in Talos. It's not part of `misc/glibc`. So the move to self-contain fabricmanager is not involved in this decision. It's purely because newly introduced NVLSM needs it

frezbo: I wonder what brings in `libgcc_s.so`

frezbo: > And that's why I introduced Wolfi here, because it's versionned and make for a stable build

this is not true, while wolfi image is pinned, any `apk add` would bring in a new version of package.

frezbo: We really need to test this in CI with the DGX/HGX machines

nvidia-gpu/nvidia-fabricmanager/lts/pkg.yaml
frezbo: ahh I see, missed this was wolfi base

nvidia-gpu/nvidia-fabricmanager/lts/pkg.yaml
npdgm: > We really need to test this in CI with the DGX/HGX machines

Agreed. All I can tell you is it runs fine in production for Blackwell DGX with the current state of this PR applied to 1.10.4

npdgm: > this is not true, while wolfi image is pinned, any `apk add` would bring in a new version of package.

Oh.. of course I can see why now. I'll dig deeper into the build system to find a solution

frezbo:
> Is this likely to merge for 1.11.0 or are you in a freeze window already? I guess it wouldn't hit a patchlevel later. No pressure, I just need a little insight to adjust ops planning with clients. Thanks!

if we can sort out everything before https://github.com/siderolabs/talos/issues/10907 beta.0 it we can pull this in

frezbo:
@npdgm is this good for another round of review?

frezbo:
> > @npdgm is this good for another round of review?
> 
> Yes! It's been in use with multi-GPU inference in production for 3 months, no problem. Lately built on 1.10.4. I still can't validate this on pre-Blackwell baseboards myself, sorry. Also any new changes would take some time to test on B200 as I would need to schedule a maintenance. Resources are scarce...

Cool, I'll get this worked on and run our tests and see if this works for existing stuff

nvidia-gpu/nvidia-fabricmanager/nvidia-fabricmanager-wrapper/pkg.yaml
```diff
@@ -0,0 +1,52 @@
+name: nvidia-fabricmanager-wrapper
+variant: scratch
+shell: /bin/bash
+dependencies:
+  - image: cgr.dev/chainguard/wolfi-base@{{ .WOLFI_BASE_REF }}
+install:
+  - bash
+  - build-base
```
frezbo, line 8: we need to lock down gcc and go versions, sometimes these updates used to break nvidia-container-toolkit, see the existing examples

nvidia-gpu/nvidia-fabricmanager/lts/pkg.yaml
```diff
@@ -2,56 +2,93 @@ name: nvidia-fabricmanager-lts
 variant: scratch
 shell: /bin/bash
 dependencies:
- - stage: base
+  - stage: base
```
frezbo, line 5: i think we don't need to do this hack anymore? since this stage doesn't seem to use anything from wolfi anymore?

nvidia-gpu/nvidia-fabricmanager/production/pkg.yaml
```diff
@@ -2,50 +2,85 @@ name: nvidia-fabricmanager-production
...
       - url: https://developer.download.nvidia.com/compute/nvidia-driver/redist/fabricmanager/linux-x86_64/fabricmanager-linux-x86_64-{{ .NVIDIA_DRIVER_PRODUCTION_VERSION }}-archive.tar.xz
         destination: fabricmanager.tar.xz
         sha256: {{ .NVIDIA_FABRIC_MANAGER_PRODUCTION_AMD64_SHA256 }}
         sha512: {{ .NVIDIA_FABRIC_MANAGER_PRODUCTION_AMD64_SHA512 }}
-    # {{ end }} This in fact is YAML comment, but Go templating instruction is evaluated by bldr
+      - url: https://developer.download.nvidia.com/compute/cuda/redist/nvlsm/linux-x86_64/nvlsm-linux-x86_64-{{ .NVIDIA_NVLSM_VERSION }}-archive.tar.xz
+        destination: nvlsm.tar.xz
+        sha256: {{ .NVIDIA_NVLSM_AMD64_SHA256 }}
+        sha512: {{ .NVIDIA_NVLSM_AMD64_SHA512 }}
+    # {{ end }} This in fact is YAML comment, but Go templating instruction is evaluated by bld
```
dsseng: nit

```suggestion
    # {{ end }} This in fact is YAML comment, but Go templating instruction is evaluated by bldr
```

nvidia-gpu/nvidia-fabricmanager/lts/pkg.yaml
```diff
@@ -2,56 +2,93 @@ name: nvidia-fabricmanager-lts
...
+    from: /
+    to: /base-rootfs
+  - image: cgr.dev/chainguard/wolfi-base@{{ .WOLFI_BASE_REF }}
```
dsseng: Agreed, this shouldn't need Wolfi IIUC, since no Glibc build happens here

nvidia-gpu/nvidia-fabricmanager/lts/pkg.yaml
npdgm: Fixed.

Sorry I had it fixed in `production` but not `lts`. Both are the same now, except for vars.
So the base stage is back to `/`, and Wolfi is pulled to `/wolfi-base`. We still need Wolfi because of the `libgcc_s.so.1` file we copy to the extension rootfs, for NVLSM. No package is installed so that file is locked down by `WOLFI_BASE_REF`.  Anyway, GCC's runtime library is quite small and portable, we could pick one from any distro or version and that would work.

nvidia-gpu/nvidia-fabricmanager/lts/pkg.yaml
npdgm: Thanks. See my answer to frezbo above, Wolfi is moved to it's own root but still needed, as a source for GCC's runtime library that's not present in Talos.

frezbo: Let's create a meta `pkg` and copy the file to the rootfs in there, then use that as depends on here, so the intent is clear and obvious

nvidia-gpu/nvidia-fabricmanager/nvidia-fabricmanager-wrapper/pkg.yaml
npdgm: Got it, I replicated your changes from #719

nvidia-gpu/nvidia-fabricmanager/lts/pkg.yaml
npdgm: It looks much better indeed. Let me know what you think I wasn't sure how to name it.
I'm satisfied with how it builds so far. Once this round of review is cleared I'll schedule an upgrade for validation.

frezbo: this looks good, could you validate from your side? Then will see about getting this merged and run through our NVIDIA tests

npdgm: sure, I got a testing slot for today. FYI this will be a cherry-pick of 778abf1feba082c3b59599ef894bd9d15c0c1db8, [pkgs#7b1183b](https://github.com/siderolabs/pkgs/commit/7b1183bea84e46cd8f1a775f95683b8a0039c2d7), and [talos#38c4ce4](https://github.com/siderolabs/talos/commit/38c4ce415dc8535b4a7403f7a35c5440f2f4aeb6) on top of v1.10.6.

npdgm: It failed with filesystem issues and I had to rollback the test machine to my 1.10.4 build. Same problem after a reset and fresh install, it's not just the upgrade.
Since 1.10.4 my build script changed a lot so perhaps there's an issue there...

```
NODE         SERVICE                    STATE     HEALTH   LAST CHANGE   LAST EVENT
10.x.x.x   apid                       Running   OK       38s ago       Health check successful
10.x.x.x   auditd                     Running   OK       50s ago       Health check successful
10.x.x.x   containerd                 Running   OK       50s ago       Health check successful
10.x.x.x   cri                        Running   OK       38s ago       Health check successful
10.x.x.x   dashboard                  Running   ?        40s ago       Process Process(["/sbin/dashboard"]) started with PID 14099
10.x.x.x   ext-nvidia-fabricmanager   Failed    ?        38s ago       Failed to create runner: mkdir /usr/local/glibc: read-only file system
10.x.x.x   kubelet                    Waiting   Fail     4s ago        Error running Containerd(kubelet), going to restart forever: task "kubelet" failed: exit code 1
10.x.x.x   machined                   Running   OK       50s ago       Health check successful
10.x.x.x   syslogd                    Running   OK       49s ago       Health check successful
10.x.x.x   udevd                      Running   OK       41s ago       Health check successful
```

kubelet
```
10.x.x.x: Flag --kube-reserved has been deprecated, This parameter should be set via the config file specified by the Kubelet's --config flag. See https://kubernetes.io/docs/tasks/administer-cluster/kubelet-config-file/ for more information.
10.x.x.x: Flag --system-reserved has been deprecated, This parameter should be set via the config file specified by the Kubelet's --config flag. See https://kubernetes.io/docs/tasks/administer-cluster/kubelet-config-file/ for more information.
10.x.x.x: I0814 11:06:57.933107   15121 codec.go:100] "Using lenient decoding as strict decoding failed" err="strict decoding error: unknown field \"crashLoopBackOff\", unknown field \"logging.options.text\""
10.x.x.x: {"ts":1755169617939.6987,"caller":"app/server.go:492","msg":"Kubelet version","v":0,"kubeletVersion":"v1.29.15"}
10.x.x.x: {"ts":1755169617939.738,"caller":"app/server.go:494","msg":"Golang settings","v":0,"GOGC":"","GOMAXPROCS":"","GOTRACEBACK":""}
10.x.x.x: {"ts":1755169617939.9963,"caller":"app/server.go:924","msg":"Client rotation is on, will bootstrap in background","v":0}
10.x.x.x: {"ts":1755169617941.5178,"caller":"certificate/certificate_store.go:130","msg":"Loading cert/key pair from \"/var/lib/kubelet/pki/kubelet-client-current.pem\".","v":0}
10.x.x.x: {"ts":1755169617943.3433,"caller":"dynamiccertificates/dynamic_cafile_content.go:157","msg":"Starting controller","v":0,"name":"client-ca-bundle::/etc/kubernetes/pki/ca.crt"}
10.x.x.x: {"ts":1755169617946.7803,"caller":"cli/run.go:74","msg":"command failed","err":"failed to run Kubelet: validate service connection: validate CRI v1 runtime API for endpoint \"unix:///run/containerd/containerd.sock\": rpc error: code = Unimplemented desc = unknown servic
e runtime.v1.RuntimeService"}
```

containerd
```
10.x.x.x: {"level":"info","msg":"starting containerd","revision":"fb4c30d4ede3531652d86197bf3fc9515e5276d9","time":"2025-08-14T11:26:13.342734006Z","version":"v2.0.5"}
10.x.x.x: {"id":"io.containerd.image-verifier.v1.bindir","level":"info","msg":"loading plugin","time":"2025-08-14T11:26:13.345779660Z","type":"io.containerd.image-verifier.v1"}
10.x.x.x: {"id":"io.containerd.warning.v1.deprecations","level":"info","msg":"loading plugin","time":"2025-08-14T11:26:13.345800696Z","type":"io.containerd.warning.v1"}
10.x.x.x: {"id":"io.containerd.content.v1.content","level":"info","msg":"loading plugin","time":"2025-08-14T11:26:13.345811130Z","type":"io.containerd.content.v1"}
10.x.x.x: {"id":"io.containerd.snapshotter.v1.native","level":"info","msg":"loading plugin","time":"2025-08-14T11:26:13.345846779Z","type":"io.containerd.snapshotter.v1"}
10.x.x.x: {"id":"io.containerd.snapshotter.v1.overlayfs","level":"info","msg":"loading plugin","time":"2025-08-14T11:26:13.345873984Z","type":"io.containerd.snapshotter.v1"}
10.x.x.x: {"id":"io.containerd.event.v1.exchange","level":"info","msg":"loading plugin","time":"2025-08-14T11:26:13.346031428Z","type":"io.containerd.event.v1"}
10.x.x.x: {"id":"io.containerd.monitor.task.v1.cgroups","level":"info","msg":"loading plugin","time":"2025-08-14T11:26:13.346069265Z","type":"io.containerd.monitor.task.v1"}
10.x.x.x: {"id":"io.containerd.metadata.v1.bolt","level":"info","msg":"loading plugin","time":"2025-08-14T11:26:13.347973039Z","type":"io.containerd.metadata.v1"}
10.x.x.x: {"level":"info","msg":"metadata content store policy set","policy":"shared","time":"2025-08-14T11:26:13.348000749Z"}
10.x.x.x: {"id":"io.containerd.gc.v1.scheduler","level":"info","msg":"loading plugin","time":"2025-08-14T11:26:13.348465032Z","type":"io.containerd.gc.v1"}
10.x.x.x: {"id":"io.containerd.differ.v1.walking","level":"info","msg":"loading plugin","time":"2025-08-14T11:26:13.348486244Z","type":"io.containerd.differ.v1"}
10.x.x.x: {"id":"io.containerd.lease.v1.manager","level":"info","msg":"loading plugin","time":"2025-08-14T11:26:13.348496446Z","type":"io.containerd.lease.v1"}
10.x.x.x: {"id":"io.containerd.service.v1.containers-service","level":"info","msg":"loading plugin","time":"2025-08-14T11:26:13.348502996Z","type":"io.containerd.service.v1"}
10.x.x.x: {"id":"io.containerd.service.v1.content-service","level":"info","msg":"loading plugin","time":"2025-08-14T11:26:13.348509565Z","type":"io.containerd.service.v1"}
10.x.x.x: {"id":"io.containerd.service.v1.diff-service","level":"info","msg":"loading plugin","time":"2025-08-14T11:26:13.348517456Z","type":"io.containerd.service.v1"}
10.x.x.x: {"id":"io.containerd.service.v1.images-service","level":"info","msg":"loading plugin","time":"2025-08-14T11:26:13.348524159Z","type":"io.containerd.service.v1"}
10.x.x.x: {"id":"io.containerd.service.v1.introspection-service","level":"info","msg":"loading plugin","time":"2025-08-14T11:26:13.348531379Z","type":"io.containerd.service.v1"}
10.x.x.x: {"id":"io.containerd.service.v1.namespaces-service","level":"info","msg":"loading plugin","time":"2025-08-14T11:26:13.348544331Z","type":"io.containerd.service.v1"}
10.x.x.x: {"id":"io.containerd.service.v1.snapshots-service","level":"info","msg":"loading plugin","time":"2025-08-14T11:26:13.348556139Z","type":"io.containerd.service.v1"}
10.x.x.x: {"id":"io.containerd.shim.v1.manager","level":"info","msg":"loading plugin","time":"2025-08-14T11:26:13.348561599Z","type":"io.containerd.shim.v1"}
10.x.x.x: {"id":"io.containerd.runtime.v2.task","level":"info","msg":"loading plugin","time":"2025-08-14T11:26:13.348571833Z","type":"io.containerd.runtime.v2"}
10.x.x.x: {"id":"io.containerd.service.v1.tasks-service","level":"info","msg":"loading plugin","time":"2025-08-14T11:26:13.348630444Z","type":"io.containerd.service.v1"}
10.x.x.x: {"id":"io.containerd.grpc.v1.containers","level":"info","msg":"loading plugin","time":"2025-08-14T11:26:13.348639425Z","type":"io.containerd.grpc.v1"}
10.x.x.x: {"id":"io.containerd.grpc.v1.content","level":"info","msg":"loading plugin","time":"2025-08-14T11:26:13.348648384Z","type":"io.containerd.grpc.v1"}
10.x.x.x: {"id":"io.containerd.grpc.v1.diff","level":"info","msg":"loading plugin","time":"2025-08-14T11:26:13.348658507Z","type":"io.containerd.grpc.v1"}
10.x.x.x: {"id":"io.containerd.grpc.v1.events","level":"info","msg":"loading plugin","time":"2025-08-14T11:26:13.348667973Z","type":"io.containerd.grpc.v1"}
10.x.x.x: {"id":"io.containerd.grpc.v1.images","level":"info","msg":"loading plugin","time":"2025-08-14T11:26:13.348673637Z","type":"io.containerd.grpc.v1"}
10.x.x.x: {"id":"io.containerd.grpc.v1.introspection","level":"info","msg":"loading plugin","time":"2025-08-14T11:26:13.348688464Z","type":"io.containerd.grpc.v1"}
10.x.x.x: {"id":"io.containerd.grpc.v1.leases","level":"info","msg":"loading plugin","time":"2025-08-14T11:26:13.348694700Z","type":"io.containerd.grpc.v1"}
10.x.x.x: {"id":"io.containerd.grpc.v1.namespaces","level":"info","msg":"loading plugin","time":"2025-08-14T11:26:13.348700846Z","type":"io.containerd.grpc.v1"}
10.x.x.x: {"id":"io.containerd.grpc.v1.snapshots","level":"info","msg":"loading plugin","time":"2025-08-14T11:26:13.348712088Z","type":"io.containerd.grpc.v1"}
10.x.x.x: {"id":"io.containerd.streaming.v1.manager","level":"info","msg":"loading plugin","time":"2025-08-14T11:26:13.348718903Z","type":"io.containerd.streaming.v1"}
10.x.x.x: {"id":"io.containerd.grpc.v1.streaming","level":"info","msg":"loading plugin","time":"2025-08-14T11:26:13.348724673Z","type":"io.containerd.grpc.v1"}
10.x.x.x: {"id":"io.containerd.grpc.v1.tasks","level":"info","msg":"loading plugin","time":"2025-08-14T11:26:13.348730302Z","type":"io.containerd.grpc.v1"}
10.x.x.x: {"id":"io.containerd.transfer.v1.local","level":"info","msg":"loading plugin","time":"2025-08-14T11:26:13.348740000Z","type":"io.containerd.transfer.v1"}
10.x.x.x: {"id":"io.containerd.grpc.v1.transfer","level":"info","msg":"loading plugin","time":"2025-08-14T11:26:13.348753786Z","type":"io.containerd.grpc.v1"}
10.x.x.x: {"id":"io.containerd.grpc.v1.version","level":"info","msg":"loading plugin","time":"2025-08-14T11:26:13.348759977Z","type":"io.containerd.grpc.v1"}
10.x.x.x: {"id":"io.containerd.ttrpc.v1.otelttrpc","level":"info","msg":"loading plugin","time":"2025-08-14T11:26:13.348764818Z","type":"io.containerd.ttrpc.v1"}
10.x.x.x: {"id":"io.containerd.grpc.v1.healthcheck","level":"info","msg":"loading plugin","time":"2025-08-14T11:26:13.348769585Z","type":"io.containerd.grpc.v1"}
10.x.x.x: {"id":"io.containerd.podsandbox.controller.v1.podsandbox","level":"info","msg":"loading plugin","time":"2025-08-14T11:26:13.348775680Z","type":"io.containerd.podsandbox.controller.v1"}
10.x.x.x: {"error":"unable to init client for podsandbox: failed to get \"io.containerd.sandbox.store.v1\" plugin: no plugins registered for io.containerd.sandbox.store.v1: plugin: not found","id":"io.containerd.podsandbox.controller.v1.podsandbox","level":"warning","msg":"failed
 to load plugin","time":"2025-08-14T11:26:13.349036722Z","type":"io.containerd.podsandbox.controller.v1"}
10.x.x.x: {"address":"/system/run/containerd/containerd.sock.ttrpc","level":"info","msg":"serving...","time":"2025-08-14T11:26:13.349202944Z"}
10.x.x.x: {"address":"/system/run/containerd/containerd.sock","level":"info","msg":"serving...","time":"2025-08-14T11:26:13.349233793Z"}
10.x.x.x: {"level":"info","msg":"containerd successfully booted in 0.008917s","time":"2025-08-14T11:26:13.349244131Z"}
10.x.x.x: {"address":"unix:///run/containerd/s/814eeaf75962b52d1228c72642b2a1e34f6ad18299621a2bc0a0d887e86f9db6","level":"info","msg":"connecting to shim apid","namespace":"system","protocol":"ttrpc","time":"2025-08-14T11:26:25.618484895Z","version":3}
```

npdgm: Ok that was a completely unrelated issue with [talhelper](https://github.com/budimanjojo/talhelper).
We have been using this talhelper config for a while:
```yaml
    machineFiles:
      - op: create
        content: |
          [plugins]
            [plugins."io.containerd.cri.v1.runtime"]
              [plugins."io.containerd.cri.v1.runtime".containerd]
                default_runtime_name = "nvidia"
        path: /etc/cri/conf.d/20-customization.part
```
Turns out it has been generating a `machine.file:` entry with `permissions: 0o0`, which is bad but used to work. Now it breaks on 1.10.6, so `permissions: 0o444` has to be set explicitly with talhelper.

frezbo: so is this good to go?

npdgm: Can't tell yet, sorry. Last maintenance window was wasted on the issue with containerd file permission.
I've got a new slot from my client, we'll upgrade nodes on Tueday at 11 GMT.

npdgm: It failed again. I don't get it, looks like I can't make working builds anymore.

Previously I never had to call the imager with `--system-extension-image ghcr.io/siderolabs/glibc:x`. Now if I don't, then the service container fails to bind mount `/usr/local/glibc` because it doesn't exist.
But then, when explicitely imaging with the glibc extention, then I still get errors I had never see before about libraries such as :
```
nv-fabricmanager: /usr/bin/nv-fabricmanager: error while loading shared libraries: libz.so.1: cannot open shared object file: No such file or directory
```

frezbo: if glibc is added as a dependency it should work, maybe it's looking for `libz` from the alpine base

npdgm: Sorry about the noise, I figured it out. Silly problem on my side.

So it's good to go, I validated the current PR head.
- rebased on 1.10.6 : OK and now used in production
- rebased on release-1.11 : OK too

nvidia-gpu/vars.yaml
```diff
@@ -19,12 +24,17 @@ NVIDIA_FABRIC_MANAGER_PRODUCTION_ARM64_SHA256: f631a5753e6cee12978eab8e63db5b9e6
 NVIDIA_FABRIC_MANAGER_PRODUCTION_ARM64_SHA512: 6606cb088f055e4511c94cdd772bdf3f0ae118b8cb5f60533351b587e058d2fdb34078ff7aac6e96db2c2f6d993f69229f1d8ed655c8e8a29c4d2992d545694d
 NVIDIA_FABRIC_MANAGER_PRODUCTION_AMD64_SHA256: 8d24cacde4554d471899ad426f46a349d5ca0a2e8acd45c2a76381c8f496491e
 NVIDIA_FABRIC_MANAGER_PRODUCTION_AMD64_SHA512: f893f92e144c46d2f2c114399c11873ab843bfec480652878453a07ec3a8ed9af429937ca33f60039e120512a47a75ca032f0de7d37f8a8ebf385d90a0099007
+NVIDIA_NVLSM_PRODUCTION_VERSION: 2025.03.1
```
frezbo, line 27: how is the production and lts versions determined?

frezbo (approved):

frezbo (approved):

PR #739, opened 2025-06-15
feat: add DVBSky S952 firmware driver extension

yehia2amer:
feat: enhance dvb-m88ds3103 extension with IR support configuration

This system extension provides the dvb-demod-m88ds3103.fw firmware for DVB-S/S2 PCIe cards like DVBSky S952

### DIFF
--- a/dvb/dvb-m88ds3103/pkg.yaml
+++ b/dvb/dvb-m88ds3103/pkg.yaml
@@ -1,0 +1,43 @@
+name: dvb-m88ds3103
+variant: scratch
+shell: /bin/bash
+dependencies:
+  - stage: base
+  - stage: dvb-cx23885
+  # - image: alpine:3.18 # Using alpine as a base image
+  #   to: / # Ensure it copies to the root
+  # The pkgs version for a particular release of Talos as defined in
+  # https://github.com/siderolabs/talos/blob/<talos version>/pkg/machinery/gendata/data/pkgs
+  - image: "{{ .BUILD_ARG_PKGS_PREFIX }}/kernel:{{ .BUILD_ARG_PKGS }}"
+steps:
+  - sources:
+      - url: https://github.com/OpenELEC/dvb-firmware/raw/master/firmware/dvb-demod-m88ds3103.fw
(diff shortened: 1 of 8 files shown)


yehia2amer:
@smira Could you please review this?

dvb/dvb-m88ds3103/pkg.yaml
```diff
@@ -0,0 +1,43 @@
+name: dvb-m88ds3103
+variant: scratch
+shell: /bin/bash
+dependencies:
+  - stage: base
+  - stage: dvb-cx23885
```
frezbo, line 7: the `dvb-cx23885` is already copied here, so I guess there's no need to say in docs to include that as dependency

dvb/dvb-m88ds3103/pkg.yaml
```diff
@@ -0,0 +1,43 @@
...
+        # Create modprobe.d directory and cx23885.conf for IR support
+        mkdir -p /rootfs/etc/modprobe.d
+        echo "options cx23885 enable_885_ir=1" > /rootfs/etc/modprobe.d/cx23885.conf
+    test:
+      - |
+        mkdir -p /extensions-validator-rootfs
+        cp -r /rootfs/ /extensions-validator-rootfs/rootfs
+        cp /pkg/manifest.yaml /extensions-validator-rootfs/manifest.yaml
+        /extensions-validator validate --rootfs=/extensions-validator-rootfs --pkg-name="${PKG_NAME}"
+    # {{ end }} This in fact is YAML comment, but Go templating instruction is evaluated by bldr
```
frezbo: we also need an `else` clause to ship an empty extension for arm64

dvb/dvb-m88ds3103/pkg.yaml
yehia2amer: Docs Updated.

dvb/dvb-m88ds3103/pkg.yaml
yehia2amer: Added

frezbo (approved):

PR #740, opened 2025-06-22
feature: Fossorial Newt client extension

campfred:
Hi! I've been working on adding support for [Fossorial's Newt client](https://docs.fossorial.io/Newt/overview).
It's basically CloudflareD but self-hosted and open source.

It's the first time I've built an extension to a platform like Talos.
So, if anything needs to be corrected, let me know!


network/newt/newt.yaml
```diff
@@ -0,0 +1,12 @@
+name: cloudflared
```
frezbo: this needs updating

network/newt/pkg.yaml
```diff
@@ -0,0 +1,42 @@
+name: newt
+variant: scratch
+shell: /bin/bash
+dependencies:
+  - stage: base
+steps:
+  - sources:
+      # {{ if eq .ARCH "aarch64" }} This in fact is YAML comment, but Go templating instruction is evaluated by bldr
+      - url: https://github.com/fosrl/newt/releases/download/{{ .NEWT_VERSION }}/newt_linux_arm64
```
frezbo: we can just do a `make local` from source, also makes sure the binary is static

network/newt/newt.yaml
campfred: Good catch! I updated it in commit 4ad4ccb31a255b4b5562692259b55ebeeffbe76d.

network/newt/pkg.yaml
campfred: I'm not sure I understood correctly.
Do you mean we should pull the Newt git repo and build it from source instead of pulling the binary?

frezbo: yes, this is also fine as long as the published binaries are statically linked, but building it here makes sure that it's built right always

campfred: Good idea. I'll try implementing that.

campfred: I implemented building from source in ea936f3667034641f999961c22a3a7a47a53a4ef. It was easier than I anticipated.
Hopefully that's better!

network/newt/pkg.yaml
```diff
@@ -0,0 +1,54 @@
...
+    prepare:
+      - |
+        sed -i 's#$NEWT_VERSION#{{ .NEWT_VERSION }}#' /pkg/manifest.yaml
+      - |
+        mkdir newt
+        tar -xzvf newt.tar.gz --strip-components=1 -C newt
+      - | # https://github.com/fosrl/newt?tab=readme-ov-file#binary
+        cd newt
+        if [ -f main.go ]; then
+          sed -i 's/Newt version replaceme/Newt version '"{{ .NEWT_VERSION }}"'/' main.go
```
frezbo, line 27: is this needed, I couldn't find reference to version in main.go in the git repo here: https://github.com/fosrl/newt/blob/main/main.go

campfred: It is based on their own instructions for building the binary in the link at the line on top of this one.
If we don't do that, it will say `Newt version replaceme` instead of `Newt version 1.2.1` for example when calling `newt --version`.

campfred: Btw, what you're looking for is at line 403 in the `main.go` file you linked me.

frezbo: ahh I see, though I don't like how they do it, it's fine for now, I would have expected they used a variable and set that with linker flags

frezbo (approved):

network/newt/pkg.yaml
campfred: Agreed!

PR #752, opened 2025-07-04
fix: TLS Verification failures on Newt

### LINKED ISSUE #665: Tailscale extension can't talk to its API
Hi,

I installed the tailscale system extension. Some features require it to talk to its API, for example, configuring it with an oauth-style tokens:

```
apiVersion: v1alpha1
kind: ExtensionServiceConfig
name: tailscale
environment:
  - TS_AUTHKEY=tskey-client-<oauth-client-authkey>
```

This fails when it attempts to communicate with Tailscales HTTPS-secured API:
```
k8s-node1: Post "https://api.tailscale.com/api/v2/tailnet/-/keys": Post "https://api.tailscale.com/api/v2/oauth/token": tls: failed to verify certificate: x509: certificate signed by unknown authority
k8s-node1: boot: 2025/04/14 04:41:33 failed to auth tailscale: failed to auth tailscale: tailscale up failed: exit status 1
```

I suspect it needs CA root certificates, a proposed change might be to add the following mount to https://github.com/siderolabs/extensions/blob/main/network/tailscale/tailscale.yaml as is done in other containers: 
```
 - source: /etc/ssl/certs
      destination: /etc/ssl/certs
      type: bind
      options:
        - rbind
        - ro
```

campfred:
This was an oversight from me. I didn't test it with a secure endpoint when I wrote that extension.

It fixes errors like this in the extension's logs.
```log
talos.example.ca: ERROR: 2025/07/04 16:33:50 Failed to connect: failed to get token: failed to request new token: Post "https://pangolin.example.ca/api/v1/auth/newt/get-token": tls: failed to verify certificate: x509: certificate signed by unknown authority. Retrying in 10s...
```

It is based on the work done on Tailscale in #669 which fixed #665.
I figured Tailscale was a good example since it needs to reach a public endpoint that's of course using secure connections too.

### DIFF
--- a/network/newt/newt.yaml
+++ b/network/newt/newt.yaml
@@ -9,4 +9,12 @@ depends:
   - configuration: true
 container:
   entrypoint: /usr/local/bin/newt
+  mounts:
+  # Mount CA root certificates as the endpoints will be signed by Let's Encrypt
+  - source: /etc/ssl/certs
+    destination: /etc/ssl/certs
+    type: bind
+    options:
+      - rbind
+      - ro
 restart: always


smira:
Can you please sign-off on the commit? (`git commit -s`)

`conform/commit/gpg*` can be ignored

campfred:
Done! ✅

frezbo (approved):

PR #753, opened 2025-07-06
feat: add nfs-utils extension

buroa:
This builds `nfs-utils` which provides a particular utility called `nfsrahead`. It also creates the necessary udev rule to take effect on nfs mounts.

https://man7.org/linux/man-pages/man5/nfsrahead.5.html


storage/nfs-utils/pkg.yaml
```diff
@@ -0,0 +1,43 @@
+name: nfs-utils
+variant: alpine
```
frezbo: this should probably use our `base`

Makefile
```diff
@@ -90,6 +90,7 @@ TARGETS += mei
 TARGETS += metal-agent
 TARGETS += nebula
 TARGETS += newt
+TARGETS += nfs-utils
```
frezbo: this needs to be added to `.kres.yaml` and run a `make rekres`

storage/nfs-utils/pkg.yaml
```diff
@@ -0,0 +1,43 @@
...
+        destination: nfs-utils.tar.xz
+        sha256: {{ .NFS_UTILS_SHA256 }}
+        sha512: {{ .NFS_UTILS_SHA512 }}
+    prepare:
+      - |
+        tar -xf nfs-utils.tar.xz --strip-components=1
+
+        export PKG_CONFIG_PATH=/usr/lib/pkgconfig
+
+        ./configure \
```
frezbo: can the tool be compiled to look for config file under `/usr/local/etc`? Then it makes it easier to add that via an extensionserviceconfig

storage/nfs-utils/pkg.yaml
buroa: It is pretty hard to get this to build using scratch/base. Reading over the `configure` script, it always needs deps like `sqlite-devel`, no matter how many flags I disable. Do we really want to build that?

storage/nfs-utils/pkg.yaml
```diff
@@ -0,0 +1,50 @@
...
+      PKG_CONFIG_PATH: /usr/local/lib/pkgconfig
+    prepare:
+      - |
+        tar -xf nfs-utils.tar.xz --strip-components=1
+
+        mkdir -p /usr/local/include/sys
+        cp /pkg/patches/sys-queue.h /usr/local/include/sys/queue.h
+
+        ./configure \
+          --disable-gss \
```
frezbo: since this is disabling everything else, is this only used as a binary in udev rule?

buroa: That's one of the main reasons I built it, but it still has many other binaries bundled as well.

frezbo: are those needed?

buroa: > are those needed?

No, not for what I need here. I removed everything else and only install the nfsrahead util inside the extension now.

frezbo:
> > doesn't the extension needs the libevent and sqlite shared objects
> > ```shell
> > tree _out/nfsrahead/
> > _out/nfsrahead/
> > └── rootfs
> >     └── usr
> >         ├── lib
> >         │   └── udev
> >         │       └── rules.d
> >         │           └── 99-nfs.rules
> >         ├── libexec
> >         │   └── nfsrahead
> >         └── local
> >             └── etc
> >                 └── nfsrahead
> >                     └── nfs.conf
> > ```
> 
> Pretty sure the nfsrahead tool doesn't need anything like that.
> 
> https://fossies.org/linux/nfs-utils/tools/nfsrahead/main.c

super weird then it needs libevent and sqlite as build deps

buroa:
> super weird then it needs libevent and sqlite as build deps

The upstream configure script is all over the place, so it still requires things to run it even though they aren't needed to make.

frezbo:
> > super weird then it needs libevent and sqlite as build deps
> 
> The upstream configure script is all over the place, so it still requires things to run it even though they aren't needed to make.

it doesn't make any sense to build these then, just add a patch to configure to skip these

frezbo:
> > I looked into it, it's a configure hell as you said, let's keep as is
> 
> It wasn't hard to change the section that needed the modification. I updated the PR.

it's fine, don't want to keep maintaining the patch

PR #756, opened 2025-07-15
feat: add SBOM to some extensions

dsseng:
Tailscale is built here, with SBOM generated in its pkg.yaml, and
nvidia-lts copies SBOM from pkgs.

Signed-off-by: Dmitrii Sharshakov <dmitry.sharshakov@siderolabs.com>

### DIFF
--- a/network/tailscale/pkg.yaml
+++ b/network/tailscale/pkg.yaml
@@ -42,6 +42,13 @@ steps:
       - |
         mkdir -p /rootfs/usr/local/etc/containers
         cp /pkg/tailscale.yaml /rootfs/usr/local/etc/containers/
+    sbom:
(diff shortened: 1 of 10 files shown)


network/tailscale/pkg.yaml
```diff
@@ -42,6 +42,13 @@ steps:
       - |
         mkdir -p /rootfs/usr/local/etc/containers
         cp /pkg/tailscale.yaml /rootfs/usr/local/etc/containers/
+    sbom:
+      outputPath: /rootfs/usr/local/share/spdx/tailscale.spdx.json
```
dsseng, line 46: This will need a change in Talos to find extensions' SBOMs, will probably do that a bit later

smira (approved):

PR #758, opened 2025-07-17
fix: build of nfsrahead extension

smira:
It was completely broken:

* missing validator
* missing manifest
* incorrect version
* wrong install prefix


storage/nfsrahead/pkg.yaml
```diff
@@ -46,8 +49,16 @@ steps:
       - |
         mkdir -p /rootfs/usr/local/etc/nfsrahead
         cp /pkg/files/nfs.conf /rootfs/usr/local/etc/nfsrahead/nfs.conf
-        make -C tools/nfsrahead DESTDIR=/rootfs install
+        make -C tools/nfsrahead DESTDIR=/rootfs  install
```
dsseng: nit (perfectionism): extra whitespace

smira: yes, thanks!

PR #762, opened 2025-07-23
docs: add SBOM for container-runtimes

shanduur:
Partial for #10940


container-runtime/crun/pkg.yaml
```diff
@@ -34,7 +34,14 @@ steps:
         cp -r /rootfs/ /extensions-validator-rootfs/rootfs
         cp /pkg/manifest.yaml /extensions-validator-rootfs/manifest.yaml
         /extensions-validator validate --rootfs=/extensions-validator-rootfs --pkg-name="${PKG_NAME}"
-
+    sbom:
+      outputPath: /rootfs/usr/share/spdx/crun.spdx.json
```
smira: I _think_ it should be under `/rootfs/usr/local` for extensions

shanduur: Fixed (for reference, the paths are defined here: https://github.com/siderolabs/talos/blob/85d6b91/pkg/machinery/constants/constants.go#L1268-L1272)

smira: The way I noticed that is the fact that extensions have limited paths they can "write" to, and `/usr/share` is not included: https://github.com/siderolabs/talos/blob/85d6b919890a1aa9c4f94d5b18861cc617134ff9/pkg/machinery/extensions/extensions.go#L11-L27

I think the way SBOM generation works, it "skips" the validator which should have warned about it

frezbo: interesting, so this happens after the rootfs is copied?

smira: @dsseng should have more details on this one, I think it runs after all build steps

container-runtime/ecr-credential-provider/pkg.yaml
```diff
@@ -45,6 +45,12 @@ steps:
         cp -r /rootfs/ /extensions-validator-rootfs/rootfs
         cp /pkg/manifest.yaml /extensions-validator-rootfs/manifest.yaml
         /extensions-validator validate --rootfs=/extensions-validator-rootfs --pkg-name="${PKG_NAME}"
+    sbom:
+      outputPath: /rootfs/usr/local/share/spdx/ecr-credential-provider.spdx.json
+      version: {{ .CLOUD_PROVIDER_AWS_VERSION }}
+      cpes: []
```
dsseng: Can be dropped entirely iirc.

container-runtime/crun/pkg.yaml
dsseng: It runs after steps, but before finalize I believe. Best way to be sure: inspect the build result, extensions' SBOMs must be in /usr/local/share/spdx.

And it does skip the validator, yeah. Would be happy if anyone suggests how do I make it run after all steps but `test`

smira: my only idea is to make it a "step" of its own, so we can re-arrange the validator to be the step after sbom

dsseng: that would likely make the build more complex and might lower the performance

Issue #530, opened 2024-11-24
thunderbolt-udev extension


onedr0p:
I'm trying to get my sonnet thunderbolt NICs working in Talos maintenance mode so I can use them to bootstrap Talos. However they aren't recognized by Talos until Talos installed via the following config with the existing thunderbolt extension.

```yaml
version: v1alpha1
machine:
...
  udev:
    rules:
      - ACTION=="add", SUBSYSTEM=="thunderbolt", ATTR{authorized}=="0", ATTR{authorized}="1"
  kernel:
    modules:
      - name: thunderbolt
      - name: thunderbolt_net
...
```

@frezbo mentioned it should be possible to create a new system extension that can add that udev rule in order for the NICs to be recognized in maintenance mode.

Another option would be to include the udev rule in the tb extension.

onedr0p:
Still a problem

PR #763, opened 2025-07-23
feat: add thunderbolt udev rule

buroa:
This will authorize devices on system boot

Fixes: https://github.com/siderolabs/extensions/issues/530

### DIFF
--- a/drivers/thunderbolt/README.md
+++ b/drivers/thunderbolt/README.md
@@ -66,3 +66,7 @@ You can also verify everything in dmesg:
  SUBSYSTEM=thunderbolt
  DEVICE=+thunderbolt:1-1
 ```
+
+## Security Warning
+
+This extension automatically authorizes all Thunderbolt devices during system boot, which poses potential security risks. Use at your own discretion.
--- a/drivers/thunderbolt/files/99-thunderbolt.rules
+++ b/drivers/thunderbolt/files/99-thunderbolt.rules
@@ -1,0 +1,2 @@
+# This will authorize Thunderbolt devices on system boot
+ACTION=="add", SUBSYSTEM=="thunderbolt", ATTR{authorized}=="0", ATTR{authorized}="1"
--- a/drivers/thunderbolt/manifest.yaml
+++ b/drivers/thunderbolt/manifest.yaml
@@ -5,7 +5,8 @@ metadata:
   author: Igor Rzegocki
   description: |
     This system extension provides Thunderbolt/USB4 drivers kernel modules built against a specific Talos version.
-    This driver enables Thunderbolt/USB4 devices, including networking.
+    It enables support for Thunderbolt/USB4 devices, including those used for networking.
+    WARNING: This extension automatically authorizes all Thunderbolt devices during system boot, which poses potential security risks. Use at your own discretion.
   compatibility:
     talos:
       version: ">= v1.5.0"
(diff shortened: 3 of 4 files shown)


drivers/thunderbolt/files/99-thunderbolt.rules
```diff
@@ -0,0 +1,2 @@
+# This will authorize Thunderbolt devices on system boot
+ACTION=="add", SUBSYSTEM=="thunderbolt", ATTR{authorized}=="0", ATTR{authorized}="1"
```
smira, line 2: that I guess doesn't work great from security perspective?

buroa: I ever so slightly agree, however, do we want to build a new extension just for this?

smira: udev rules can be injected via the machine config iirc

I think if someone having Thunderbolt, and installs this extension even not using Thunderbolt exposes them to security risks immediately.

buroa: Yes, but not during install. @onedr0p, chime in here.

smira: What is the production usecase for Thunderbolt even? 

I wonder if we should change the extension description to be: "don't use it in production if concerned about physical system security".

onedr0p: @smira The issue is I cannot get access to my thunderbolt NICs unless Talos is installed. Meaning they are not recognized in maintenance mode.

buroa: > What is the production usecase for Thunderbolt even?
> 
> I wonder if we should change the extension description to be: "don't use it in production if concerned about physical system security".

I agree. I have never used Thunderbolt in production (work).

smira: What is the hardware that needs Thunderbolt NICs? 

I'm fine with making it "skip authorization" if the extension's description clearly states the security risks.

buroa: > What is the hardware that needs Thunderbolt NICs?
> 
> I'm fine with making it "skip authorization" if the extension's description clearly states the security risks.

Typically to upgrade the network of a node; you can just grab a thunderbolt nic to add 10G, for example.

buroa: @smira Updated the description of the extension. Hopefully it's clear.

onedr0p: These are the devices I have plugged into some Intel NUCs, they provide me 10Gb networking. https://www.sonnettech.com/product/solo10g-sfp-tb3/overview.html

drivers/thunderbolt/README.md
```diff
@@ -66,3 +66,7 @@ You can also verify everything in dmesg:
  SUBSYSTEM=thunderbolt
  DEVICE=+thunderbolt:1-1
 ```
+
```
frezbo, line 69: I think an ideal solution would be to have a new talos machine config document for udev rules that gets applied even in maintenance, wdyt @smira

smira: yes, that would certainly be the best approach, but I guess we are not there yet...

smira (approved):

drivers/thunderbolt/files/99-thunderbolt.rules
smira: thank you!

PR #764, opened 2025-07-23
docs: add SBOM for more extensions

shanduur:
Partial for #10940


storage/drbd/pkg.yaml
```diff
@@ -21,6 +21,14 @@ steps:
         cp -r /rootfs/ /extensions-validator-rootfs/rootfs
         cp /pkg/manifest.yaml /extensions-validator-rootfs/manifest.yaml
         /extensions-validator validate --rootfs=/extensions-validator-rootfs --pkg-name="${PKG_NAME}"
+    sbom:
+      outputPath: /rootfs/usr/local/share/spdx/drbd.spdx.json
+      version: {{ .DRBD_DRIVER_VERSION }}
+      cpes:
+        - cpe:2.3:a:linbit:drbd8:{{ .DRBD_DRIVER_VERSION }}:*:*:*:*:*:*:*
+        - cpe:2.3:a:linbit:drbd9:{{ .DRBD_DRIVER_VERSION }}:*:*:*:*:*:*:*
```
shanduur: `linbit:drbd8` exists, but looks like it's exclusively for `drbd` v8.x.y. We can assume that new vulnerability can be published under `linbit:drbd9`, but I am not sure if it should be put here.

CC @dsseng

tools/ctr/pkg.yaml
```diff
@@ -17,6 +17,12 @@ steps:
       cp -r /rootfs/ /extensions-validator-rootfs/rootfs
       cp /pkg/manifest.yaml /extensions-validator-rootfs/manifest.yaml
       /extensions-validator validate --rootfs=/extensions-validator-rootfs --pkg-name="${PKG_NAME}"
+    sbom:
+      outputPath: /rootfs/usr/local/share/spdx/ctr.spdx.json
+      version: {{ .VERSION }}
+      purl: pkg:golang/github.com/containerd/containerd@{{ .VERSION }}
```
shanduur: PURL here, as ctr uses client libs and is part of containerd?

storage/drbd/pkg.yaml
dsseng: looks good, that's their fault of not following the CPE spec

guest-agents/metal-agent/pkg.yaml
```diff
@@ -24,6 +24,11 @@ steps:
         cp -r /rootfs/ /extensions-validator-rootfs/rootfs
         cp /pkg/manifest.yaml /extensions-validator-rootfs/manifest.yaml
         /extensions-validator validate --rootfs=/extensions-validator-rootfs --pkg-name="${PKG_NAME}"
+    sbom:
+      outputPath: /rootfs/usr/local/share/spdx/metal-agent.spdx.json
+      version: {{ .VERSION }}
+      licenses:
+        - MPL-2.0
```
dsseng, line 31: This is a Go project: https://github.com/siderolabs/talos-metal-agent

Maybe it's worth generating an SBOM in its Dockerfile, just like it's done for Talos. Not sure. @smira wdyt?

storage/drbd/pkg.yaml
```diff
@@ -21,6 +21,14 @@ steps:
         cp -r /rootfs/ /extensions-validator-rootfs/rootfs
         cp /pkg/manifest.yaml /extensions-validator-rootfs/manifest.yaml
         /extensions-validator validate --rootfs=/extensions-validator-rootfs --pkg-name="${PKG_NAME}"
+    sbom:
+      outputPath: /rootfs/usr/local/share/spdx/drbd.spdx.json
+      version: {{ .DRBD_DRIVER_VERSION }}
+      cpes:
+        - cpe:2.3:a:linbit:drbd8:{{ .DRBD_DRIVER_VERSION }}:*:*:*:*:*:*:*
```
dsseng: Let's hope they'll adhere to the proper format with version being in the dedicated field

```suggestion
        - cpe:2.3:a:linbit:drbd:{{ .DRBD_DRIVER_VERSION }}:*:*:*:*:*:*:*
        - cpe:2.3:a:linbit:drbd8:{{ .DRBD_DRIVER_VERSION }}:*:*:*:*:*:*:*
```

dsseng: Oh, by the way. Here we copy a package from pkgs, which should have an own SBOM. As we do basically nothing new here, no added components, we should build an SBOM in pkgs (look at NVIDIA non-free modules as an example, their SBOM is also copied in the extensions repo), so that it captures data about what source files we downloaded.

tools/ctr/pkg.yaml
dsseng: I guess. Containerd itself is covered by Talos core SBOM, and golang dependency is in there as well. Let's keep it like this.

storage/zfs/pkg.yaml
```diff
@@ -27,6 +27,13 @@ steps:
         cp -r /rootfs/ /extensions-validator-rootfs/rootfs
         cp /pkg/manifest.yaml /extensions-validator-rootfs/manifest.yaml
         /extensions-validator validate --rootfs=/extensions-validator-rootfs --pkg-name="${PKG_NAME}"
+    sbom:
+      outputPath: /rootfs/usr/local/share/spdx/zfs.spdx.json
```
dsseng: Again, here we only copy modules built in pkgs, so it's handled like NVIDIA nonfree: build an SBOM in pkgs, copy to the appropriate location here. Maybe it'd be more complex if a wrapper is added here (like nvidia-fabric-manager here, which has a Go-based wrapper process)

dsseng (commented):
Please consider moving kernel modules' SBOMs to pkgs, so that they have information on downloaded sources.

guest-agents/metal-agent/pkg.yaml
smira: I don't think we need a lot here, this agent only runs in a special mode of metal provider, and never on Talos in "production" mode, so nobody would see this SBOM for real

dsseng: Okay, can be skipped then

tools/ctr/pkg.yaml
frezbo: but we ship a binary?

dsseng: It will probably be covered by either containerd binary or go module, both of which are there in Talos main

frezbo: ahh I see since ctr is part of containerd

smira: we could probably just pick up SBOM from pkgs/containerd - it literally just extracts a single binary

storage/zfs/pkg.yaml
shanduur: Removed.

storage/drbd/pkg.yaml
shanduur: Removed.

tools/ctr/pkg.yaml
shanduur: Removed SBOM for `ctr`.

storage/drbd/pkg.yaml
dsseng: Should we postpone introducing SBOMs for these components? I think they are pretty valuable to have, as those 1) have real CPEs and have had known vulnerabilities, 2) are kernel drivers, thus pretty important for system security. So it'd be great to not drop them, but move to pkgs and do the copying here

shanduur: I'd prefer to move without those SBOMs for now, and add them later in `pkgs`?

dsseng: Yes, let's go this way

PR #765, opened 2025-07-23
fix: nfsrahead udev rule

buroa:
Upstream has the `99-nfs.rules` artifact already built, so we need to run a `make clean` to delete this file and recreate it with the correct prefix: `/usr/local`. Otherwise, the udev rule will never work because it's pointing to this binary `/usr/libexec/nfsrahead` instead of the generated one `/usr/local/libexec/nfsrahead`

+ https://github.com/siderolabs/extensions/commit/c479d91284a4f98236306d6efce1ab8a46d43a74

### DIFF
--- a/storage/nfsrahead/pkg.yaml
+++ b/storage/nfsrahead/pkg.yaml
@@ -29,28 +29,29 @@ steps:
         export CFLAGS="${CFLAGS} -I/usr/local/include"
 
         ./configure \
-          --disable-gss \
-          --disable-ipv6 \
-          --disable-mount \
-          --disable-nfsdcld \
-          --disable-nfsdctl \
-          --disable-nfsv4 \
-          --disable-nfsv41 \
-          --disable-uuid \
-          --enable-junction=no \
-          --with-nfsconfig=/usr/local/etc/nfsrahead/nfs.conf \
-          --with-rpcgen=internal \
-          --prefix=/usr/local
+            --disable-gss \
+            --disable-ipv6 \
+            --disable-mount \
+            --disable-nfsdcld \
+            --disable-nfsdctl \
+            --disable-nfsv4 \
+            --disable-nfsv41 \
+            --disable-uuid \
+            --enable-junction=no \
+            --with-nfsconfig=/usr/local/etc/nfsrahead/nfs.conf \
+            --with-rpcgen=internal \
+            --prefix=/usr/local
     build:
       - |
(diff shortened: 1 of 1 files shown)


storage/nfsrahead/pkg.yaml
```diff
@@ -29,28 +29,29 @@ steps:
...
+            --disable-nfsv4 \
+            --disable-nfsv41 \
+            --disable-uuid \
+            --enable-junction=no \
+            --with-nfsconfig=/usr/local/etc/nfsrahead/nfs.conf \
+            --with-rpcgen=internal \
+            --prefix=/usr/local
     build:
       - |
+        make clean
```
frezbo, line 46: I'm not sure how this would fix things, the `99` rules is in talos rootfs, not here, so a proper fix would be to either drop it from `pkgs` like as in https://github.com/siderolabs/pkgs/blob/main/systemd-udevd/pkg.yaml#L33 or patch it there

buroa: The .rules is in the nfs-utils tar.gz. Because it's already generated, running a make on it won't recreate it. However, running a make clean does delete it and then uses the .in file to regenerate it correctly.

https://git.linux-nfs.org/?p=steved/nfs-utils.git;a=tree;f=tools/nfsrahead;h=d2184c6deb0c5fa9b6a97d0a731cbe4944149c37;hb=HEAD

```
|-- manifest.yaml
`-- rootfs
    `-- usr
        |-- lib
        |   `-- udev
        |       `-- rules.d
        |           `-- 99-nfs.rules
        `-- local
            |-- etc
            |   `-- nfsrahead
            |       `-- nfs.conf
            `-- libexec
                `-- nfsrahead
```

---

Before *without* `make clean`:
```
SUBSYSTEM=="bdi", ACTION=="add", PROGRAM="/usr/libexec/nfsrahead %k", ATTR{read_ahead_kb}="%c"
```

After adding `make clean`:
```
SUBSYSTEM=="bdi", ACTION=="add", PROGRAM="/usr/local/libexec/nfsrahead %k", ATTR{read_ahead_kb}="%c"
```

Notice the program path changed, and now it's correct because the nfsrahead executable is found at `/usr/local/libexec/nfsrahead`.

smira (approved):
lgtm, waiting for @frezbo review

frezbo (approved):

PR #767, opened 2025-07-25
feat(ci): add grype scan

frezbo:
Add Grype scan in CI.


internal/grype-scan/pkg.yaml
```diff
@@ -0,0 +1,117 @@
...
+      - /root/.cache/grype/db
+    build:
+      - |
+        mkdir -p /rootfs
+        touch /rootfs/grype-scan.log
+        cd /pkg
+
+        go tool -modfile=go.mod \
+        github.com/anchore/grype/cmd/grype \
+        sbom:/rootfs/combined.spdx -vv 2>&1 | tee /rootfs/grype-scan.log
```
dsseng:
```suggestion
        sbom:/rootfs/combined.spdx --fail-on negligible -vv 2>&1 | tee /rootfs/grype-scan.log
```

This would make it fail in case any vulnerability is discovered. But due to quite some false-positives this might need to be postponed to after those are dealt with

internal/grype-scan/pkg.yaml
```diff
@@ -0,0 +1,117 @@
...
+      - |
+        mkdir -p /rootfs
+        cd /pkg
+
+        go tool -modfile=go.mod \
+        github.com/anchore/syft/cmd/syft \
+        scan --from dir /usr/local/share/spdx \
+        --select-catalogers "+sbom-cataloger" \
+        --source-name "Talos extensions" --source-version {{ .BUILD_ARG_TAG }} \
+        -o spdx-json > /rootfs/combined.spdx
```
dsseng:
```suggestion
        -o spdx-json > /rootfs/combined.spdx.json
```

So various tools like SBOM viewer can identify it more easily

dsseng (approved):

PR #770, opened 2025-07-31
docs: add more extension SBOMs

shanduur:
Status by this PR:

  - Has sbom and cpes or purl: `18 files`
  - Has sbom, but no cpes/purl: `18 files`
  - No sbom: `56 files`
      - References source by url: `13 files`

Partial for https://github.com/siderolabs/talos/issues/10940

<details>

```bash
#!/usr/bin/env bash

# Output files for each category
category_a="_out/sbom.txt"
category_b="_out/sbom-incomplete.txt"
category_c="_out/sbom-missing.txt"
category_d="_out/sbom-missing-has-source.txt"

# Clear previous output
> "$category_a"
> "$category_b"
> "$category_c"
> "$category_d"

# Find all pkg.yaml files
find . -type f -name "pkg.yaml" ! -path "*/internal/*" | while read -r file; do
    # Check if file contains "sbom" key
    if grep -qE '^\s*sbom\s*:' "$file"; then
        # Now check for cpes or purl keys
        if grep -qE '^\s*(cpes|purl)\s*:' "$file"; then
            echo "$file" >> "$category_a"  # Category A
        else
            echo "$file" >> "$category_b"  # Category B
        fi
    else
        echo "$file" >> "$category_c"      # Category C

        if grep -q "url: https://" "$file"; then
            echo "$file" >> "$category_d"
        fi
    fi
done

# Summary output
echo "Categorization complete:"
echo "  - Has sbom and cpes or purl: \`$(wc -l < "$category_a") files\`"
echo "  - Has sbom, but no cpes/purl: \`$(wc -l < "$category_b") files\`"
echo "  - No sbom: \`$(wc -l < "$category_c") files\`"
echo "      - References source by url: \`$(wc -l < "$category_d") files\`"
```

</details>

### DIFF
--- a/dvb/cx23885/pkg.yaml
+++ b/dvb/cx23885/pkg.yaml
@@ -46,6 +46,9 @@ steps:
         cp -r /rootfs/ /extensions-validator-rootfs/rootfs
         cp /pkg/manifest.yaml /extensions-validator-rootfs/manifest.yaml
         /extensions-validator validate --rootfs=/extensions-validator-rootfs --pkg-name="${PKG_NAME}"
+    sbom:
+      outputPath: /rootfs/usr/local/share/spdx/cx23885.spdx.json
+      version: {{ .LINUX_DVB_FIRMWARE }}
     # {{ end }} This in fact is YAML comment, but Go templating instruction is evaluated by bldr
 finalize:
   - from: /rootfs
--- a/guest-agents/qemu-guest-agent/pcre2/pkg.yaml
+++ b/guest-agents/qemu-guest-agent/pcre2/pkg.yaml
@@ -23,6 +23,13 @@ steps:
     install:
       - |
         make DESTDIR=/rootfs install
+    sbom:
+      outputPath: /rootfs/usr/local/share/spdx/pcre2.spdx.json
+      version: {{ .PCRE2_VERSION }}
+      cpes:
+        - cpe:2.3:a:pcre:pcre2:{{ .CONTAINER_TOOLKIT_VERSION }}:*:*:*:*:*:*:*
+      licenses:
+        - BSD-2-Clause
 finalize:
   - from: /rootfs
     to: /
--- a/storage/zfs/zfs-tools/pkg.yaml
+++ b/storage/zfs/zfs-tools/pkg.yaml
@@ -14,7 +14,7 @@ steps:
         sha256: {{ .ZFS_TOOLS_SHA256 }}
         sha512: {{ .ZFS_TOOLS_SHA512 }}
     env:
-      ARCH: {{ if eq .ARCH "aarch64"}}arm64{{ else if eq .ARCH "x86_64" }}x86_64{{ else }}unsupported{{ end }}
+      ARCH: '{{ if eq .ARCH "aarch64"}}arm64{{ else if eq .ARCH "x86_64" }}x86_64{{ else }}unsupported{{ end }}'
     prepare:
       - |
         tar -xf zfs.tar.gz --strip-components=1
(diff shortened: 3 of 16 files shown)


smira (approved):

dvb/cx23885/pkg.yaml
```diff
@@ -46,6 +46,9 @@ steps:
         cp -r /rootfs/ /extensions-validator-rootfs/rootfs
         cp /pkg/manifest.yaml /extensions-validator-rootfs/manifest.yaml
         /extensions-validator validate --rootfs=/extensions-validator-rootfs --pkg-name="${PKG_NAME}"
+    sbom:
+      outputPath: /rootfs/usr/local/share/spdx/cx23885.spdx.json
+      version: {{ .LINUX_DVB_FIRMWARE }}
```
dsseng, line 51: These all seem to be part of Linux + part of linux-firmware. We track Linux as a whole in the core Talos

guest-agents/qemu-guest-agent/pcre2/pkg.yaml
```diff
@@ -23,6 +23,13 @@ steps:
     install:
       - |
         make DESTDIR=/rootfs install
+    sbom:
+      outputPath: /rootfs/usr/local/share/spdx/pcre2.spdx.json
+      version: {{ .PCRE2_VERSION }}
+      cpes:
+        - cpe:2.3:a:pcre:pcre2:{{ .CONTAINER_TOOLKIT_VERSION }}:*:*:*:*:*:*:*
```
dsseng, line 30:
```suggestion
        - cpe:2.3:a:pcre:pcre2:{{ .PCRE2_VERSION }}:*:*:*:*:*:*:*
```

Also not sure if it should be built here, it's already being built in pkgs. But that'd be a separate topic for a PR

storage/zfs/zfs-tools/pkg.yaml
```diff
@@ -40,6 +40,13 @@ steps:
         # since talos doesn't have a shell, this would fail
         # the default shipped file is a no-op, so safe to remove.
         rm -f /rootfs/usr/local/libexec/zfs/zfs_prepare_disk
+    sbom:
+      outputPath: /rootfs/usr/local/share/spdx/zfs-tools.spdx.json
+      version: {{ .ZFS_DRIVER_VERSION }}
+      cpes:
+        - cpe:2.3:a:openzfs:openzfs:{{ .ZFS_DRIVER_VERSION }}:*:*:*:*:*:*:*
```
dsseng, line 47: Don't we already have the same entry in the kernel module build in pkgs? Maybe this shouldn't be duplicated then, but not so sure

dsseng (approved):
LGTM apart from some nits

storage/zfs/zfs-tools/pkg.yaml
smira: This also ships `zfs` CLI in addition to kernel module

dsseng: Yes, but should it ship another SBOM with the same data? How would scanners see that potentially?

smira: idk, we can leave zfs SBOM here, as it's more than just a kernel module

frezbo: i think we should wrap the one from zfs kernel module and the one from zfs cli in the extension sbom, we can fix once bldr has support, I'll take a look into bldr

shanduur:
Ah, GitHub lagged for some reason and showed comments after I merged 🤦🏻 I’ll push fix.

PR #814, opened 2025-09-03
docs: Update ZFS README for existing pools

MaximUltimatum:
Adds documentation for troubleshooting existing pools


MaximUltimatum:
@kvaps @donch 

I thought this info should go somewhere in the documentation, and this seemed like an ok place to add it?

This zpool came from a rocky server and Cockpit added those SELinux flags. When I moved the server to Talos, it took a while to figure out why my child datasets weren't mounting. Having a note here like this would have helped me. Happy to move the info somewhere else if you don't think it's a good fit here.

storage/zfs/README.md
```diff
@@ -122,3 +122,22 @@ talosctl mounts | grep <mount of zfs raid you specified (/var/...)>
...
+
+## Use existing pools
+
+If you have existing pools, they will be automatically detected and used.
+
+However, you may need to enter a debug shell and modify the pool configuration in order for them to work with talos.
+Turn off samba and nfs share settings.
+In addition, some SELinux pool policies are currently incompatible with talos. If you see errors like this:
+```bash
+node_name: kern: warning: [2025-09-03T00:34:28.924235026Z]: SELinux: Unable to set superblock options before the security server is initialized
```
dsseng, line 134: Did you explicitly disable SELinux (e.g. by passing `selinux=0`)? What version of Talos did you find this issue?

By default, Talos loads a policy in permissive mode. `selinux=0` disables the SELinux LSM, and `enforcing=1` makes the policy enforced (not supported nor tested with extensions enabled).

This seems to be a somewhat rare warning, does this prevent ZFS volumes from being mounted?

MaximUltimatum: Yes I explicitly disabled SELinux by setting selinux=0. I tested on 1.11.0 and 1.10.7 both with and without selinux=0. I was not enforcing the policy.

It does prevent ZFS volumes from being mounted. On startup, Talos will mount the volume, however, `mount -a` will return an error  for child datasets with unsupported SELinux policies set. It was something like "pool_name/child_dataset_name invalid argument". The child dataset folders will be visible under `/var/mnt/pool_name` but will be empty.

dsseng: If it fails without selinux=0, then I guess it could potentially be caused by missing labels. Talos does not have SELinux labels coming from a Rocky install, so you'll have to disable those via volume options anyway I guess.

I wonder if a volume that has SELinux on, but no foreign labels mounts fine.

But so far this step seems to make sense

storage/zfs/README.md
```diff
@@ -122,3 +122,22 @@ talosctl mounts | grep <mount of zfs raid you specified (/var/...)>
...
 
 You should see the mount.
+
+## Use existing pools
+
+If you have existing pools, they will be automatically detected and used.
+
+However, you may need to enter a debug shell and modify the pool configuration in order for them to work with talos.
+Turn off samba and nfs share settings.
+In addition, some SELinux pool policies are currently incompatible with talos. If you see errors like this:
```
dsseng:
```suggestion
However, you may need to enter a debug shell and modify the pool configuration in order for them to work with Talos.
Turn off Samba and NFS share settings.
In addition, some SELinux pool policies are currently incompatible with Talos. If you see errors like this:
```

storage/zfs/README.md
```diff
@@ -122,3 +122,22 @@ talosctl mounts | grep <mount of zfs raid you specified (/var/...)>
...
+```bash
+node_name: kern: warning: [2025-09-03T00:34:28.924235026Z]: SELinux: Unable to set superblock options before the security server is initialized
+```
+You can resolve this by modifying the zfs pool settings:
```
dsseng:
```suggestion
You can resolve this by modifying the ZFS pool settings to disable SELinux on the volume:
```

MaximUltimatum:
Do I need to upload my gpg public key somewhere so the signature check passes? Haven't had to do that before. 

Also I just squashed and force-pushed so the suggestions I accepted wouldn't fail commit body checks, but it looks like I'm somehow 407 commits ahead of main... I'm not sure I understand how that happened.

MaximUltimatum:
Just confirming this isn't waiting on me. If helpful, I could provide a sanitized talos config and ZFS configuration to reproduce the issue.

dsseng:
Seems to be in review. Not blocked on you I guess

frezbo (approved):

PR #815, opened 2025-09-03
feat: update dependencies

smira:
crun: 1.23.1
kata-containers: 3.20.0
spin: 0.21.0
stargz-snapshotter: 0.17.0
youki: 0.5.5
nebula: 1.9.6
newt: 1.4.4
tailscale: 1.86.2
fuse3: 3.17.4
nut-client: 2.8.4


misc/glibc/pkg.yaml
```diff
@@ -15,7 +15,7 @@ steps:
       CC: gcc-14
       CXX: g++-14
     sources:
-      - url: https://src.fedoraproject.org/lookaside/pkgs/glibc/glibc-{{ .GLIBC_VERSION }}-8-g1e0e33e1b1.tar.xz/sha512/449e3d4f6b59bfde2175c5d1be71447b084e0b12b176518b65fd9b0ac8430766b25416a173fe3efd47462bc1719d59a051e7eed9544e0fba9165dd86f69ee0b9/glibc-{{ .GLIBC_VERSION }}-8-g1e0e33e1b1.tar.xz
+      - url: https://src.fedoraproject.org/lookaside/pkgs/glibc/glibc-{{ .GLIBC_VERSION }}-21-g7a8f3c6ee4.tar.xz/sha512/526a26116174517b4a35a9cbe74fca7c092728992f54fccfb54ee853eb4217553d8c4fe91117e3a27894f265dd4f63975c7fe85c561de4922b83c0594073a097/glibc-{{ .GLIBC_VERSION }}-21-g7a8f3c6ee4.tar.xz
```
frezbo: let's also make the sha512 use the go template variable

smira: I dropped glibc update, as it fails to build

frezbo (approved):

misc/glibc/pkg.yaml
frezbo: probably gcc update, I was building when I tried the update for the grype scan

smira: it fails in a weird way, exit code is 1, but there is no clear error message even without `-j`

PR #822, opened 2025-09-09
Add beszel-agent extension for monitoring talos nodes

virtualm2000:
Before migrating to talos all my nodes has beszel agents to monitor the system metrics. With this extension I can run the agent again.


smira:
Can you run it as a daemonset in Kubernetes?

smira:
> Yes, it's possible, but running it independent from the Kubernetes cluster, sometimes gives you more information of what is happening.

What is the actual difference?

PR #823, opened 2025-09-14
[feat]: add usb-audio-drivers

mindstorms6:
I would like to enable usb audio on one of my nodes - this follows the `usb-modem-drivers` on how to enable that. I have built and published the extension here: https://github.com/users/mindstorms6/packages/container/package/usb-audio-drivers

I think we will ultimately need to revert this though: https://github.com/siderolabs/pkgs/pull/435 or have a different image we can copy the relevant modules from.


mindstorms6:
I see someone has already re-enabled this in the kernel. I don't think this will be needed any more.
Closing.

https://github.com/siderolabs/pkgs/commit/e067c2015302c5069676727f047b6aef9ce0dc0c

frezbo (approved):
would you also mind adding to the `MAINTAINERS` file

PR #838, opened 2025-09-29
Netbird

batleforc:
Hi,
The actual version didn't work for me no matter how I installed it. Every time I ended up with the usual message of netbird when you pass no argument. But even after fixing that, I had a further problem.
In order to make it work, I had to provide some change:
- Mount /etc/ssl/certs or else it didn't recognize netbird's certificate
- Mount /etc/os-release to allow netbird to detect the os
- Provide netbird with a "mimic" of uname has it also need it to gather metadata on the host
- Make sure that the log is printed in the console in order to watch it with `talosctl logs ext-netbird`
- Set the daemon sock in /var/run/netbird a folder with RW (same for the config file)


network/netbird/src/main.go
frezbo, line 1: I'm not sure about adding all these to just make a tool work, it could directly query the kernel, also how was this tested in the first place if this was broken a lot 🤔

batleforc: In fact it just need to have a fake uname to work has its gard coded
Netbird need this kind of data to apply security policy

frezbo: yeh, it's just feels a bit bad, could you confirm netbird works with these changes?

network/netbird/netbird.yaml
```diff
@@ -9,6 +9,17 @@ depends:
...
   entrypoint: /usr/local/bin/netbird
+  args:
+    - "up"
+    - "--foreground-mode"
+  environment:
+    - NB_DAEMON_ADDR=unix:///var/run/netbird/netbird.sock
+    - NB_LOG_FILE=console,/var/log/netbird/client.log
+    - NB_DISABLE_PROFILES=true
+    - USER=talos
+    - NB_CONFIG=/var/run/netbird/config.json
```
frezbo, line 20: who's providing this config file?

batleforc: Netbird, that's where he will store it once he succesfuly connect

network/netbird/src/main.go
batleforc: I can and tested it  with multiple full cluster (standalone node, all node of the cluster, only one of them) i can provide how i tester the all node and standalone one

frezbo: I don't think uname is needed https://github.com/netbirdio/netbird/blob/e7b5537dcc280384470668f461bbb1f7d2f41218/client/system/info_linux.go#L81 it just logs and moves on

batleforc: no it does not https://github.com/netbirdio/netbird/blob/e7b5537dcc280384470668f461bbb1f7d2f41218/client/system/info_linux.go#L38, it then try to split by whitespace and if the data is not present it panic

frezbo: that shouldn't panic, probably not handled somewhere else, I would suggest create an upstream issue

batleforc: It should if the command does not exist, if i don't misunderstand those infos is needed in the use of the product

batleforc: For example, in netbird I can do something called a posture check that would allow me to block peer if I detect a security flaw in one of the version. That's also why i didn't make an uname that report fake data

batleforc:
Should be good @frezbo

PR #845, opened 2025-10-09
feat: add xe extension

mikesmitty:
With the upcoming 6.18 kernel Intel is enabling SR-IOV in the xe driver for most recent iGPUs (except Meteor/Arrow/Lunar Lake) behind a force_probe arg, so it would be nice to have the driver available. This is a mostly direct copy of the i915 extension so I left the author and tier as is

### DIFF
--- a/drm/xe/pkg.yaml
+++ b/drm/xe/pkg.yaml
@@ -1,0 +1,37 @@
+name: xe
+variant: scratch
+shell: /bin/bash
+dependencies:
+  - stage: base
+  # The pkgs version for a particular release of Talos as defined in
+  # https://github.com/siderolabs/talos/blob/<talos version>/pkg/machinery/gendata/data/pkgs
+  - image: "{{ .BUILD_ARG_PKGS_PREFIX }}/kernel:{{ .BUILD_ARG_PKGS }}"
+  - image: "{{ .BUILD_ARG_PKGS_PREFIX }}/linux-firmware:{{ .BUILD_ARG_PKGS }}"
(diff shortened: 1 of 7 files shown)


drm/xe/pkg.yaml
```diff
@@ -0,0 +1,38 @@
...
+        mkdir -p /rootfs
+  # {{ if eq .ARCH "x86_64" }} This in fact is YAML comment, but Go templating instruction is evaluated by bldr
+  - install:
+      - |
+        export KERNELRELEASE=$(find /usr/lib/modules -type d -name "*-talos" -exec basename {} \+)
+
+        xargs -a /pkg/files/modules.txt -I {} install -D /usr/lib/modules/${KERNELRELEASE}/{} /rootfs/usr/lib/modules/${KERNELRELEASE}/{}
+      - |
+        mkdir -p /rootfs/usr/lib/firmware
+        cp -R -p /usr/lib/firmware/i915 /rootfs/usr/lib/firmware
```
smira: (do we need both firmware directories here?)

mikesmitty: In hindsight, it's probably more proper to leave it out. The xe driver will need the i915 firmware for unsupported platforms if it's force-enabled, but anyone force-enabling it can also add the i915 extension and force-disable the i915 module with the `xe.force_probe=* i915.force_probe=!*` kernel args.

blake-hamm:
Really appreciate the add! Looking forward to it being available in the image factory

PR #854, opened 2025-10-25
feat: add the ability to have multiple zerotier network to join

Hugome:
Hello,

Simple addition to be able to join multiple ZT networks.
It will not break the actual "one network" behavior, since `strings.Split` with a separator and no occurrence found will return an array of 1 with the full string in it. 

Thanks


smira:
Forked to #859, as it's a fork from the org, and we can't push back.

PR #856, opened 2025-10-27
feat: bump deps

frezbo:
Bump dependencies.


container-runtime/vars.yaml
```diff
@@ -1,5 +1,5 @@
 # renovate: datasource=github-tags extractVersion=^release-(?<version>.*)$ depName=google/gvisor
-GVISOR_VERSION: 20251006.0
+GVISOR_VERSION: 20251020.0
```
smira, line 2: GVISOR_REV is not updated

smira (approved):

PR #857, opened 2025-10-27
NFS-utils

clementnuss:
This extension brings in [`rpcbind`](https://www.linuxfromscratch.org/blfs/view/svn/basicnet/rpcbind.html) and `rpc.statd` (from the [nfs-utils](https://www.linuxfromscratch.org/blfs/view/svn/basicnet/nfs-utils.html) tools), two daemons required to mount NFS shares with locking support.

using this extension makes it directly possible to use e.g. trident v25.06 to mount NFS shares with locks support. (https://github.com/NetApp/trident/issues/806#issuecomment-2399332314)

it would also solve the following issues: https://github.com/siderolabs/talos/issues/6582 https://github.com/siderolabs/talos/issues/8862

I set the TIER of the extension to contrib, but feel free to bump to extra if you think you want to help support the build process.

I could happily contribute some test cases, but I would require some guidance on where to implement those.


storage/zfs/zfs-tools/libtirpc/pkg.yaml
```diff
@@ -30,6 +30,9 @@ steps:
       - |
         mkdir -p /rootfs
 
+        mkdir -p /rootfs/usr/local/include/sys
```
smira, line 33: this is an extra change? (it is already done in `prepare:` phase

Makefile
```diff
@@ -95,6 +94,7 @@ TARGETS += netbird
 TARGETS += newt
 TARGETS += nfsd
 TARGETS += nfsrahead
+TARGETS += nfs-utils
```
smira: this file is automatically generated, can you please add `nfs-utils` to `.kres.yaml`?

storage/nfs-utils/rpc-statd.yaml
```diff
@@ -0,0 +1,114 @@
...
+container:
+  entrypoint: /usr/local/sbin/rpc.statd
+  args:
+    - --foreground
+    - --no-syslog
+  security:
+    maskedPaths: []
+    readonlyPaths: []
+  mounts:
+    # netconfig file required by libtirpc
```
frezbo: if these files doesn't need to be on the host, let's just put it into the container rootfs and stop using the mounts

storage/nfs-utils/rpc-statd.yaml
```diff
@@ -0,0 +1,114 @@
...
+  - network:
+      - addresses
+      - connectivity
+      - hostname
+      - etcfiles
+container:
+  entrypoint: /usr/local/sbin/rpc.statd
+  args:
+    - --foreground
+    - --no-syslog
```
frezbo, line 14: i see `/dev/log` being mounted, then why is `no-syslog`? 🤔

storage/nfs-utils/rpc-statd.yaml
clementnuss: good idea, I had not fully grasped the directory structure for extensions.

storage/nfs-utils/rpc-statd.yaml
clementnuss: good point. I confused myself. only `rpcbind` needs it (hard-coded). I've removed it

storage/zfs/zfs-tools/libtirpc/pkg.yaml
clementnuss: in the `prepare:` the folder that's created is `/usr/local/include/sys`, without the `/rootfs` prefix.
I however need the header file to be in the output of libtirpc to be able to reuse it.

clementnuss (commented):
I've addressed your comments, fixed a `musl` incompatibility, commented the patches and updated the author to PostFinance

for my part, this PR is ready. shall I squash my commits?

PR #864, opened 2025-11-05
chore: backports for v1.11.4

shanduur:


hack/release.toml
```diff
@@ -19,7 +19,7 @@ See [Talos Linux documentation](https://www.talos.dev/v1.11/talos-guides/configu
     [notes.updates]
         title = "Component Updates"
         description = """\
-linux-firmware: 20251011
+linux-firmware: 20251021
```
smira, line 22: please bump in `Pkgfile`, otherwise the version published will be wrong

smira: and this should show up in `make extensions-catalog` -> `README.md`

shanduur: Oh, so the update of firmware was skipped for 1.11.3?

smira: yes, looks like that. the actual firmware was updated, but extension version was incorrect

PR #868, opened 2025-11-06
feat: update pkgs with containerd 2.1.5

smira:
Preparing for Talos 1.11.5.

### DIFF
--- a/hack/release.toml
+++ b/hack/release.toml
@@ -6,7 +6,7 @@ github_repo = "siderolabs/extensions"
 match_deps = "^github.com/((talos-systems|siderolabs)/[a-zA-Z0-9-]+)$"
 
 # previous release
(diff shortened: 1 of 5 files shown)


hack/release.toml
```diff
@@ -19,7 +19,7 @@ See [Talos Linux documentation](https://www.talos.dev/v1.11/talos-guides/configu
     [notes.updates]
         title = "Component Updates"
         description = """\
-linux-firmware: 20251021
+ctr: 2.1.5
```
frezbo, line 22:
```suggestion
containerd: 2.1.5
```

frezbo (approved):

hack/release.toml
smira: This is extensions, so it's `ctr` here

PR #872, opened 2025-11-06
add soci snapshotter extension

zlangbert:
Adds an extension for the [AWS SOCI snapshotter](https://github.com/awslabs/soci-snapshotter). We have tested this internally at Hathora and would like to upstream the work.

I've translated our internal extension build process to the tooling in this repo, using the stargz extension as an example. I have no been able to test it locally yet.


container-runtime/soci-snapshotter/pkg.yaml
```diff
@@ -0,0 +1,69 @@
...
+        cp ./out/soci /rootfs/usr/local/lib/containers/soci-snapshotter/soci
+        chmod +x rootfs/usr/local/lib/containers/soci-snapshotter/soci
+      - |
+        mkdir -p /rootfs/etc/cri/conf.d
+        cp /pkg/10-soci-snapshotter.part /rootfs/etc/cri/conf.d/10-soci-snapshotter.part
+
+        mkdir -p /rootfs/usr/local/etc/soci-snapshotter-grpc
+        cp /pkg/config.toml /rootfs/usr/local/etc/soci-snapshotter-grpc/config.toml
+
+        mkdir -p /rootfs/usr/local/etc/containers
```
frezbo, line 54: might as well put this into the container rootfs, use use an extension service config if it needs runtime changes

zlangbert: Hey @frezbo, thanks for taking a look. Not sure I am following you comment, can you clarify? Do you mean the `config.toml` that contains the SOCI settings?

frezbo: yeh the config yaml

frezbo: it doesn't need to be on the host, right?

zlangbert: No it doesn't need to be on the host, the SOCI grpc service just needs to load it. So would the change be to just remove this config file from the extension and document a `ExtensionServiceConfig` for how to configure it?

frezbo: correct, let's just move it into container rootfs, and maybe later add the serviceconfig support so that we can land this in 1.12.

zlangbert: Done, moved the config into the container rootfs

PR #874, opened 2025-11-07
fix: glibc build for arm64

frezbo:
Also use ubuntu mirrors for sources which doesn't have `sha512sum` in the URL. `libtirpc` needed to be downgraded to 1.3.6 since ubuntu mirror doesn't have the latest yet.


Pkgfile
```diff
@@ -16,10 +16,10 @@ vars:
   TENSTORRENT_VERSION: 2.5.0 # update this when updating PKGS_VERSION in Makefile
   HAILORT_VERSION: 4.23.0 # update this when updating PKGS_VERSION in Makefile
 
-  # renovate: datasource=git-tags extractVersion=^libtiprc-(?<version>.*)$ depName=git://linux-nfs.org/~steved/libtirpc
-  LIBTIRPC_VERSION: 1-3-7
-  LIBTIRPC_SHA256: b47d3ac19d3549e54a05d0019a6c400674da716123858cfdb6d3bdd70a66c702
-  LIBTIRPC_SHA512: 50c0153bc72a3f9578eb7a9dc123ea531df83c455f19b03d2bc59563d8453fcd278a025eef05865d1218b2c9358a9152ee069ce0e4d11dfea4ca0a8a24221fad
+  # renovate: datasource=git-tags extractVersion=^libtiprc-(?<version>.*)$ depName=git://git.linux-nfs.org/projects/steved/libtirpc.git
+  LIBTIRPC_VERSION: 1-3-6
```
smira, line 20: going down on a version?

frezbo: yes, ubuntu mirrors doesn't have what we had, also the old build was using cached version, since even 1.3.6 doesn;t build without patches, so we were broken anyways, just that Pkgfile had newer versions being updated

smira (approved):

Issue #863, opened 2025-11-04
Persistent video 4 linux dev rules not preset


ian-thompson-yv:
udev rules for v4l devices were removed from the Talos `systemd-udev` pkgs in https://github.com/siderolabs/pkgs/blob/f70250f91aa43c6972ccb121e6300ace9ffa535c/systemd-udevd/pkg.yaml#L33

The `v4l-uvc` extension does not put the rules back so video devices are not mounted in the way software expects. We have written our own local extension that is just the rules and we include this alongside the v4l extension. 

Is it appropriate to raise a PR to add the new extension or should we raise a PR to modify the v4l extension and add the upstream rules (https://github.com/systemd/systemd/blob/9218e4eacc545a36a1275264013cb3d6b3eef41f/rules.d/60-persistent-v4l.rules#L4)

PR #880, opened 2025-11-13
fix: video4linux persistent udev rules

ian-thompson-yv:
The udev rules for v4l devices are removed from the Talos `systemd-udev` package in https://github.com/siderolabs/pkgs/blob/main/systemd-udevd/pkg.yaml#L33. This commit adds persistent udev rules as part of the `v4l-uvc-drivers` system extension to mount devices in the expected way.

Issue #863

### DIFF
--- a/drivers/v4l-uvc/pkg.yaml
+++ b/drivers/v4l-uvc/pkg.yaml
@@ -7,14 +7,21 @@ dependencies:
   # https://github.com/siderolabs/talos/blob/<talos version>/pkg/machinery/gendata/data/pkgs
   - image: "{{ .BUILD_ARG_PKGS_PREFIX }}/kernel:{{ .BUILD_ARG_PKGS }}"
 steps:
+  - sources:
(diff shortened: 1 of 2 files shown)


youview/yvrig-control/files/60-persistent-v4l.rules
```diff
@@ -0,0 +1,20 @@
+ACTION=="remove", GOTO="persistent_v4l_end"
```
frezbo: maybe just add this as a url souce, so we get updates from upstream if any

ian-thompson-yv: Added an upstream source from the systemd repo

ian-thompson-yv:
> @ian-thompson-yv could you DCO with `git commit -s --amend --no-edit && git push --force-with-lease`

Done, I see the failed check about my gpg signature. Can I help resolve that

Issue #576, opened 2025-01-01
[Support] Guidance / Recommendations for extension development


jabenedicic:
I am currently working on a number of additional tools/configs that would support greater storage adoption for Talos, however I am unsure how to best package these up into extensions.

Is there a recommendation that each extension should be a single tool/driver/firmware etc? Or could you bundle multiple together to fit the intended purpose?

For example let's say all tools needed for iSCSI storage arrays CSI drivers to work. This would involve open-iscsi, multipath-tools, potentially other scsi related packages. What is the best way to proceed?

Many thanks,
JB

jabenedicic:
Thanks, that helps. Will do some more testing and put together some PR's.

adamcirillo:
is there any chance you might be able to add vdo support for lvm with this?

Issue #467, opened 2024-09-10
[Request] multipath storage extension


devmechanic:
Hello,
There is few extensions in storage section but there is not one for setup multipathing. This is very important component for any SAN storage with HBAs. It provides redundancy if any path fails.
Would it be possible to create extension for it? It would be beneficial for any bare metal servers with HBAs.

Here is quick info about multipathing: https://documentation.ubuntu.com/server/explanation/intro-to/multipath/

Thank you.

Issue #134, opened 2023-03-21
[iscsi-tools] Please support multipath


Twanislas:
Hi team,

It looks like multipath-tools is not available in the iscsi-tools extension. I tried to search for multipath references everywhere, to no avail.

When mounting iscsi target trying to use multipath (using democratic-csi), I get `failed to discover multipath device`. However, I can see that the two disk references are successfully mounted :
```
{"host":"kubeworker03","level":"info","message":"new request - driver: ControllerSynologyDriver method: NodeStageVolume call: {\"metadata\":{\"x-forwarded-host\":[\"localhost\"],\"user-agent\":[\"grpc-go/1.49.0\"]},\"request\":{\"publish_context\":{},\"secrets\":\"redacted\",\"volume_context\":{\"lun\":\"1\",\"node_attach_driver\":\"iscsi\",\"portal\":\"\",\"portals\":\"nas01-iscsi0.i,nas01-iscsi1.i\",\"provisioner_driver\":\"synology-iscsi\",\"storage.kubernetes.io/csiProvisionerIdentity\":\"1679431778041-8081-synology-iscsi\",\"interface\":\"\",\"iqn\":\"iqn.2000-01.com.synology:csi.k8s-pvc-83c2d084-36e9-4b31-86f2-bca6df03770d\"},\"volume_id\":\"pvc-83c2d084-36e9-4b31-86f2-bca6df03770d\",\"staging_target_path\":\"/var/lib/kubelet/plugins/kubernetes.io/csi/synology-iscsi/4db6d710ff1e3fb3e3cd58f0a4fb57b0c962a6dd773ba289f292dbf2b48d2c61/globalmount\",\"volume_capability\":{\"access_mode\":{\"mode\":\"SINGLE_NODE_MULTI_WRITER\"},\"mount\":{\"mount_flags\":[],\"fs_type\":\"\",\"volume_mount_group\":\"\"},\"access_type\":\"mount\"}},\"cancelled\":false}","service":"democratic-csi","timestamp":"2023-03-21T21:31:11.735Z"}
executing iscsi command: iscsiadm -m node -T iqn.2000-01.com.synology:csi.k8s-pvc-83c2d084-36e9-4b31-86f2-bca6df03770d -p nas01-iscsi0.i:3260 -o new
[...]

PR #882, opened 2025-11-13
Multipath tools

Untersander:
This PR brings two new extensions [multipath-tools](https://github.com/opensvc/multipath-tools), a daemon for iSCSI devices and trident-iscsi-tools which are a bunch of executables needed by the [Trident CSI](https://github.com/NetApp/trident) to work with iSCSI. It also adds a little change in the util-linux extension for another dependency of Trident.

The multipath-tools extension brings the userspace daemon part required for SAN multipath devices to work.
The additional required multipath kernel modules are already available in Talos(or at least the ones we need).
It continues work already started in PR #505 and brings the extension to a working state without pulling in everything.
This would probably also solve the issues #576 #467 which were closed as not planned and maybe #134 as well.

The second part with trident-iscsi-tools and the change to the util-linux extension brings in binaries expected to be on the host by the Trident CSI in the case of iSCSI connections. Together these changes allow for iSCSI connections via the Trident CSI.
These binaries should come with the Trident CSI itself in my opinion and we've already explored this and will also work on a PR on the Trident side.
With this I'm not quite sure what to do, as it's really specific to Trident itself and maybe hopefully not be required sometime in the future if we can get a PR merged in Trident.

I'm also happy to move this to a second PR as the multipath-tools extension will also be useful for users of other CSIs which want to use SAN multipath devices.

I also would be glad about pointers on how do you handle the testing of extensions, so we could add them to theses as well.

Thanks to @bzub and @aaroncirillo-cision for their previous work and @SimLi1333 for the help.


storage/multipath-tools/multipathd.yaml
```diff
@@ -0,0 +1,69 @@
...
+  mounts:
+    # ld-musl-x86_64.so.1
+    - source: /lib
+      destination: /lib
+      type: bind
+      options:
+        - bind
+        - ro
+    # /etc/multipath.conf /etc/multipath/bindings
+    - source: /usr/local/lib/containers/multipathd/etc
```
frezbo: we should just copy these files into the container rootfs at `/usr/local/lib/containers/multipathd/etc` rather than doing a mount, also this mount is really problematic in since it's mounts the container rootfs used at runtime into the container again

storage/multipath-tools/multipathd.yaml
```diff
@@ -0,0 +1,69 @@
...
+        - rw
+    # libs
+    - source: /usr/local/lib/
+      destination: /usr/local/lib/
+      type: bind
+      options:
+        - bind
+        - ro
+    # /user/bin
+    - source: /usr/local/bin
```
frezbo: drop these and copy files into `/usr/local/lib/containers/multipathd` and avoid the mounts

storage/multipath-tools/multipathd.yaml
```diff
@@ -0,0 +1,69 @@
...
+        - bind
+        - ro
+    # /dev/mapper and multipath disk
+    - source: /dev
+      destination: /dev
+      type: bind
+      options:
+        - bind
+        - rw
+    - source: /sys
```
frezbo, line 39: settings `maskedPaths: []` should give `/sys` i believe, but need to confirm

storage/multipath-tools/pkg.yaml
```diff
@@ -0,0 +1,71 @@
...
+      - |
+        rm -rf /rootfs/usr/local/share
+        rm -rf /rootfs/usr/local/include
+      # This file tries to create a tmpfs mount at `/var/run/multipath`.
+      - |
+        rm /rootfs/usr/lib/tmpfiles.d/multipath.conf
+        rmdir /rootfs/usr/lib/tmpfiles.d
+      # Removed but might be needed by other users of multipath-tools
+      - |
+        rm /rootfs/usr/lib/udev/kpartx_id
```
frezbo, line 57: if it's a small binary we can keep it

storage/trident-iscsi-tools/vars.yaml
```diff
@@ -0,0 +1,7 @@
+VERSION: v0.0.1
+
+LSSCSI_VERSION: 0.32
```
frezbo, line 4: https://sg.danny.cz/scsi/lsscsi.html lists as a source for tags https://github.com/doug-gilbert/lsscsi, could use for renovate

.kres.yaml
```diff
@@ -58,6 +59,7 @@ spec:
     - tailscale
     - tenstorrent
     - thunderbolt
+    - trident-iscsi-tools
```
frezbo, line 63: ~I don't think it hurts to package this into existing iscsi tools~

storage/trident-iscsi-tools/pkg.yaml
```diff
@@ -0,0 +1,46 @@
...
+    install:
+      - |
+        make DESTDIR=/rootfs install
+      # Remove man pages and info pages to save space
+      - |
+        rm -rf /rootfs/usr/local/share
+      - |
+        mkdir -p /rootfs/usr/local/bin
+        mkdir -p /rootfs/usr/local/sbin
+        cp /usr/bin/ls /rootfs/usr/local/bin
```
frezbo, line 30: I don't like this, copying over basic tools into talos when all these can be done via code, then it makes for this to be a separate package (https://github.com/siderolabs/extensions/pull/882/files#r2525625614 ignore this comment)

SimLi1333: We agree they should be done in code, apparently trident just calls these binaries at runtime like `cat initiatorname.iscsi` instead of `os.ReadFile`, that is why we separated it in its own extension. Other question was, is it ok to just copy binaries from the build stage or would it be better building these from source, like I startet here: https://github.com/INSRapperswil/extensions/tree/trident-operator-tools/storage/trident-operator-tools

storage/multipath-tools/multipathd.yaml
Untersander: Seems like it does, removed this bind mount.

storage/multipath-tools/multipathd.yaml
Untersander: oh yes this was left over from some testing, has been removed now

storage/multipath-tools/multipathd.yaml
Untersander: The Trident CSI and probably others depend on the binaries being available on the host, because they usually are, so that they can communicate with the multipath devices (checking for state/config, flushing operations, reloading of devices, etc.). We'll try to get a fix in for Trident not to depend on the binaries being on the host (Trident depends on way to many binaries being on the host...). I don't like it but I believe in the mean time it would make sense to have them there until the CSIs don't depend on it anymore.

storage/multipath-tools/pkg.yaml
Untersander: > It's a bash script, which is not used by Trident, although I can not tell if other CSIs or tools needing multipath depend on it. The reason why I removed it is because the extensions-validator-rootfs says its not allowed at this path. Thanks for all the feedback and taking time to help us!

storage/trident-iscsi-tools/vars.yaml
SimLi1333: Based on some other examples I changed the renovate annotations

storage/trident-iscsi-tools/pkg.yaml
frezbo: since it's on the host, you can copy from `base` since it's all built from same toolchain

storage/multipath-tools/pkg.yaml
frezbo: if it's bash let's drop it

storage/multipath-tools/bindings
```diff
@@ -0,0 +1,7 @@
+# Multipath bindings, Version : 1.0
```
frezbo: doesn't the multipath source ship this? if so let's copy from it

storage/multipath-tools/patches/disable-exception-handling.patch
```diff
@@ -0,0 +1,14 @@
...
+index 729618bd..91118ec9 100644
+--- a/Makefile.inc
++++ b/Makefile.inc
+@@ -115,8 +115,7 @@ CPPFLAGS	:= $(FORTIFY_OPT) $(CPPFLAGS) $(D_URCU_VERSION) \
+ 		   -DRUNTIME_DIR=\"$(runtimedir)\" -DCONFIG_DIR=\"$(TGTDIR)$(configdir)\" \
+ 		   -DDEFAULT_CONFIGFILE=\"$(TGTDIR)$(configfile)\" -DSTATE_DIR=\"$(TGTDIR)$(statedir)\" \
+ 		   -DEXTRAVERSION=\"$(EXTRAVERSION)\" -MMD -MP
+-CFLAGS		:= -std=gnu99 $(CFLAGS) $(OPTFLAGS) $(WARNFLAGS) -pipe \
+-		   -fexceptions
++CFLAGS		:= -std=gnu99 $(CFLAGS) $(OPTFLAGS) $(WARNFLAGS) -pipe
```
frezbo: is this patch necessary?

SimLi1333: Apparently yes otherwise we get `Error loading shared library libgcc_s.so.1: No such file or directory (needed by /usr/local/sbin/multipathd)` and similar errors during multipath startup. An alternative would be to `cp /usr/lib/libgcc_s.so.1 /rootfs/usr/local/lib/containers/multipathd/usr/lib/` if this is the preferred way.

storage/multipath-tools/bindings
SimLi1333: Ah, yes. Since we do not mount to a read-only filesystem anymore multipath can create this file at startup

storage/multipath-tools/multipathd.yaml
SimLi1333: I reintroduced this again, as it caused `Failed to set recovery_tmo to 5: Read-only file system` in multipath log - It tries to write to `/sys/class/iscsi_session/` and only finds a read-only file system.

storage/multipath-tools/patches/disable-exception-handling.patch
frezbo: let's copy libgcc, that avoids having to maintain a patch

storage/multipath-tools/multipathd.yaml
```diff
@@ -0,0 +1,55 @@
...
+  security:
+    maskedPaths: []
+    readonlyPaths: []
+    writeableRootfs: true
+  entrypoint: /usr/local/sbin/multipathd
+  args:
+    - -d
+    - -s
+  mounts:
+    # ld-musl-x86_64.so.1
```
frezbo, line 12: there's an open issue to stop using this pattern https://github.com/siderolabs/extensions/issues/876, but I think it can be left out of this PR, and will do separately. If all good let's move out of the draft status

Untersander: Alright, yes from our side all is good.

frezbo:
> > @SimLi1333 @Untersander is this good to go?
> 
> Yes, we did the final validation of the libgcc change today and everything works as expected.

Okay, I'll work on getting this merged

PR #889, opened 2025-11-18
feat: backports for 1.10

frezbo:
Rekres and bump dependencies in sync with `pkgs`.


misc/glibc/vars.yaml
```diff
@@ -1,4 +1,4 @@
 # renovate: datasource=docker versioning=docker depName=cgr.dev/chainguard/wolfi-base
-WOLFI_BASE_REF: sha256:57428116d2d7c27d1d4de4103e19b40bb8d2942ff6dff31b900e55efedeb7e30
+WOLFI_BASE_REF: sha256:42012fa027adc864efbb7cf68d9fc575ea45fe1b9fb0d16602e00438ce3901b1
```
frezbo, line 2: existing wolfi image was broken for builds (probably the apk packages got updated)

PR #911, opened 2025-12-03
feat: add bird2 system-extension

xvzf:
This PR adds bird2 routing daemon as system extension, enabling Talos to natively integrate into L3 network topologies while keeping the configuration very flexible.
It's designed to be generic and support many different use-cases.

I've added an example on how this can be used in combination with Cilium, offloading the L3 datapath to a network fabric.

**Open Question:** I am unsure how the target should end-up on the kres generated Makefile. Suggestions here are appreciated!

Feedback appreciated, I'm open for changes/suggestions! Ping to @rothgar ;-)


network/vars.yaml
```diff
@@ -28,3 +28,7 @@ NEWT_SHA512: 021cb93f8616a907f5d0d2e5c2979cc28669042b2cc77322633a23e49cae48e0498
 NETBIRD_VERSION: 0.60.2
 NETBIRD_SHA256: e3ec023aff48d86ed1302708aa0522f610ec02891da8f24e6d98b9e9680bbc4e
 NETBIRD_SHA512: 8bebc0412932eb0c278749b282bae366ff032ca95c5e9032ad54e84e91b47c6f530c4f4bcf9d5843fdd4025516efdad46913f73d18976b177c2534bd7258affe
+# renovate: datasource=git-tags depName=https://gitlab.nic.cz/labs/bird.git allowedVersions=2.* versioning=semver
+BIRD2_VERSION: "2.17.2"
```
smira, line 32: 2.17.2 doesn't exist, there is 2.17.1 only...  not sure how it came together?

network/vars.yaml
```diff
@@ -28,3 +28,7 @@ NEWT_SHA512: 021cb93f8616a907f5d0d2e5c2979cc28669042b2cc77322633a23e49cae48e0498
 NETBIRD_VERSION: 0.60.2
 NETBIRD_SHA256: e3ec023aff48d86ed1302708aa0522f610ec02891da8f24e6d98b9e9680bbc4e
 NETBIRD_SHA512: 8bebc0412932eb0c278749b282bae366ff032ca95c5e9032ad54e84e91b47c6f530c4f4bcf9d5843fdd4025516efdad46913f73d18976b177c2534bd7258affe
+# renovate: datasource=git-tags depName=https://gitlab.nic.cz/labs/bird.git allowedVersions=2.* versioning=semver
```
smira, line 31: @frezbo , I think this is unparseable with our case, right? we need to limit versions other way?

network/vars.yaml
xvzf: Hm, seems the release has been not created but the tag definitely exists, contains a bug-fix & exists in their gitlab ([ref](https://gitlab.nic.cz/labs/bird/-/tree/v2.17.2?ref_type=tags))

xvzf:
@smira gonna open the PR up from my personal account tonight; haven't been aware about organizational forks work differently

network/vars.yaml
smira: At least it's not downloadable from the location you provided (we can use gitlab to download it)?

xvzf: Wth, I built this yesterday on a fresh system. Gonna check tonight what's going wrong here and amend/push a fix

smira:
> @smira gonna open the PR up from my personal account tonight; haven't been aware about organizational forks work differently

We can keep working on in #913 as well - it has some other fixes

network/vars.yaml
smira: https://bird.network.cz/ says 2.7.1 is the latest release (and no 2.7.2 in downloads). I downgraded it to 2.7.1 in #913

PR #895, opened 2025-11-25
fix(nvidia-fabricmanager): ignore interfaces with isSMDisabled flag set

npdgm:
Fixes: #893

Not all LPF interfaces are eligible to manage an NVSwitch and be used by NVLSM. We lack one check that's present in `nvidia-fabricmanager-start.sh`.

Changes:
- check for the isSMDisabled capability flag before selecting an LFP
- improve logging to dump information on all InfiniBand ports that could be of interest, to facilitate bug report processing
- better error reporting for InfiniBand code

PR #912, opened 2025-12-04
fix(nvidia-fabricmanager): memory leak in wrapper

npdgm:
Fixes missing memory deallocation introduced in #895. It's a tiny leak and only happened once at startup.

I just noticed that while going back to libibumad documentation. AFAIK there was no issue.
It's been built and tested on hardware.


smira (approved):
thank you!

PR #913, opened 2025-12-04
feat: add bird2 system-extension

smira:
Introduce bird2 routing daemon as system extension to talos linux. The bird2 configuration must be provided via an ExtensionServiceConfig and placed on /usr/local/etc/bird.conf

Forked from #911


xvzf:
As #913 is closed, bringing the discussion here; 2.17.1 seems to be the last official release, the gitlab artifact for the tag is seemingly re-generated.

If we can get the renovate config running, it's g2g for me :+1: - thx for pushing it forward @smira

network/vars.yaml
```diff
@@ -28,3 +28,7 @@ NEWT_SHA512: 021cb93f8616a907f5d0d2e5c2979cc28669042b2cc77322633a23e49cae48e0498
 NETBIRD_VERSION: 0.60.2
 NETBIRD_SHA256: e3ec023aff48d86ed1302708aa0522f610ec02891da8f24e6d98b9e9680bbc4e
 NETBIRD_SHA512: 8bebc0412932eb0c278749b282bae366ff032ca95c5e9032ad54e84e91b47c6f530c4f4bcf9d5843fdd4025516efdad46913f73d18976b177c2534bd7258affe
+# renovate: datasource=git-tags extractVersion=^v(?<version>.*)$ depName=https://gitlab.nic.cz/labs/bird.git
```
xvzf, line 31: one caveat; there's the 2.17.2 tag, which has no release artifact attached to it yet. I'll follow up on this

smira: yes, renovate will flag this, but we can't do much about it until it's fixed upstream

PR #916, opened 2025-12-05
Save Netbird config to a persistent location

shyam0904a:
By default the extensions stores the config on the /var/run folder which does not persist across reboots. This causes peers to be newly created. 

This commit changes the location to /var/lib and therefore not creating a new peer.

### DIFF
--- a/network/netbird/netbird.yaml
+++ b/network/netbird/netbird.yaml
@@ -17,7 +17,7 @@ container:
     - NB_LOG_FILE=console,/var/log/netbird/client.log
     - NB_DISABLE_PROFILES=true
     - USER=talos
-    - NB_CONFIG=/var/run/netbird/config.json
+    - NB_CONFIG=/var/lib/netbird/config.json
     - HOME=/var/run/netbird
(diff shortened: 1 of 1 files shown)


stevefan1999-personal:
@utkuozdemir need this before 1.12.0 release, mind giving it a merge?

stevefan1999-personal:
> > @utkuozdemir need this before 1.12.0 release, mind giving it a merge?
> 
> We're planning some major changes to how extensions can be configured making these stuff easier in the next release, so this might not make it

Well that means I can't use it for the time being...I need this because Netbird has built-in NAT traversal with native performance, useless way less RAM than Tailscale plus it can seamlessly integrate with my existing network topology with zero trust...

AEnterprise:
> > @utkuozdemir need this before 1.12.0 release, mind giving it a merge?
> 
> We're planning some major changes to how extensions can be configured making these stuff easier in the next release, so this might not make it

is there any update on this?

also the netbird plugin no longer works in its current form, the key no longer is passed to the container. Could this be a side effect of whatever this configuration change was? i can't find any info on that

network/netbird/netbird.yaml
AEnterprise: doesn't this bind also need to be updated then?

frezbo (approved):

PR #942, opened 2026-01-02
(WIP) feat: add aws-iam-authenticator extension

fabiomatavelli:
Adds the [aws-iam-authenticator](kubernetes-sigs/aws-iam-authenticator) to Talos.


fabiomatavelli:
@frezbo I'm still thinking if it would make sense to add this or not tbh. With the extension the process would be executed on all the nodes (worker and control-planes) as it requires the configuration to be present. So, I'm thinking if it makes sense to move forward with the extension or if providing the necessary files (certs and kubeconfig) during the bootstrap of the nodes would make more sense.

PR #950, opened 2026-01-18
Added VDUSE extension

D13410N3:
This PR adds a new extension for VDUSE kernel modules. This is needed for distributed storage systems like Vitastor that require better performance than NBD, and probably others as well.

### DIFF
--- a/drivers/vduse/manifest.yaml.tmpl
+++ b/drivers/vduse/manifest.yaml.tmpl
@@ -2,7 +2,7 @@ version: v1alpha1
 metadata:
   name: vduse
   version: "{{ .VERSION }}"
-  author: Dmitry
+  author: Dmitry Naumov
   description: |
     [{{ .TIER }}] This system extension provides VDUSE (vDPA Device in Userspace) kernel modules
     for Talos Linux. VDUSE enables userspace programs to implement virtio block devices with


D13410N3:
> > @frezbo Hello, you can check with `ghcr.io/d13410n3/talos-vduse:v1.0.0`
> 
> yup seems to be there, was there a corresponding PR to https://github.com/siderolabs/pkgs/?

Ahh, now I understand you properly, sry.
Seems like no PR in pkgs repo is required - https://github.com/siderolabs/pkgs/blob/30bc671d4be566ebf60b820edd54000616262e79/kernel/build/config-amd64#L6123

drivers/vduse/files/modules.txt
```diff
@@ -0,0 +1,6 @@
+modules.order
+modules.builtin
+modules.builtin.modinfo
+kernel/drivers/vdpa/vdpa.ko
```
smira, line 4: it's already part of base Talos: https://github.com/siderolabs/talos/blob/72fe98a06f31536454f201d703f8ae6a071235b5/hack/modules-amd64.txt#L209-L210

I don't see why we need this PR.

D13410N3: My bad - initially I didn't find module in older version of Talos. I can confirm that it exists in fresh Talos. Sorry for disruption.

PR #958, opened 2026-01-26
feat: add px-fuse module

frezbo:
See https://github.com/siderolabs/pkgs/pull/1451


storage/px-fuse/manifest.yaml.tmpl
```diff
@@ -0,0 +1,10 @@
+version: v1alpha1
+metadata:
+  name: px-fuse
+  version: "{{ .VERSION }}"
+  author: Sidero Labs
+  description: |
+    [{{ .TIER }}] This system extension provides kernel module driver for px-fuse built against a specific Talos version.
```
smira:
```suggestion
    [{{ .TIER }}] This system extension provides kernel module driver for Portworx px-fuse built against a specific Talos version.
```

(just to make it a bit more discoverable?)

Issue #946, opened 2026-01-09
multipathd ExtensionServiceConfig not mounting to multipathd service container


features-not-bugs:
Hey guys

Using a fresh install of Talos v1.12.1 with schematic id: `1186e369eaa2f70197b951f3bd309bffa068c7685b97eeaf5cb8e4bd8b962d7b`, multipathd doesn't seem to be configuring itself using the applied multipathd ExtensionServiceConfig

```
apiVersion: v1alpha1
kind: ExtensionServiceConfig
name: multipathd
configFiles:
    - content: |
        defaults {
            user_friendly_names yes
            find_multipaths yes
            path_selector "round-robin 0"
        }
      mountPath: /etc/multipath.conf
```

I'm not 100%, but from a bit of digging I believe the below line being commented out is resulting in the config file not being mounted in the multipathd container?

https://github.com/siderolabs/extensions/blob/46978a1ef24dfce54589bbe57e899319b53cdbb3/storage/multipath-tools/multipathd.yaml#L54

jsenecal:
@features-not-bugs I don't think this is an actual issue, you may want to close this

PR #960, opened 2026-01-28
fix(multipath-tools): add /run/udev mount for device enumeration

jsenecal:
## Summary

Adds the missing /run/udev mount to the multipathd container, fixing device enumeration failures.

This issue was discovered during the discussion in #946, where I was troubleshooting "spurious uevent, path not found" errors. While investigating, I found that the root cause was not the commented `configuration: true` line (which is intentional per @frezbo), but rather a missing mount.

## Problem

multipathd logs show "spurious uevent, path not found" errors for all devices, and no multipath maps are created in /dev/mapper/.

## Root Cause

multipathd uses libudev's udev_enumerate_scan_devices() during initialization (see discovery.c) to discover block devices. This function reads from /run/udev/data/ to enumerate initialized devices.

Without /run/udev mounted:
1. path_discovery() returns no devices because enumeration fails
2. The internal pathvec stays empty
3. Uevents still arrive via netlink, which works without /run/udev
4. find_path_by_dev() in main.c cannot find paths in the empty pathvec
5. This results in "spurious uevent, path not found" errors

## Investigation

I verified inside the container using talosctl:

- /etc/multipath.conf is correctly mounted with custom config
- /dev is accessible and shows the expected devices (sdc, sdd, sde, sdf)
- /sys is accessible and sysfs attributes like vendor, model, and wwid are readable
- /run/udev is missing. Only multipathd.pid and multipathd.socket exist in /run

```
$ talosctl ls /proc/<pid>/root/run/
.
multipathd.pid
multipathd.socket
```

The udev database at /run/udev/data/ is not available to the container.

## Fix

Add /run/udev as a read-only bind mount to multipathd.yaml.

## References

- https://github.com/openSUSE/multipath-tools/blob/master/libmultipath/discovery.c
- https://man7.org/linux/man-pages/man3/libudev.3.html
- Discussion in #946

### DIFF
--- a/storage/multipath-tools/multipathd.yaml
+++ b/storage/multipath-tools/multipathd.yaml
@@ -42,6 +42,13 @@ container:
       options:
         - bind
         - rw
+    # udev database for device enumeration
+    - source: /run/udev
+      destination: /run/udev
+      type: bind
+      options:
+        - bind
+        - ro
 depends:
   - network:
       - addresses


jsenecal:
Sorry for all the force-pushes, I'm trying to satisfy the conform checks

frezbo (approved):

Issue #876, opened 2025-11-09
Make extension services self contained


frezbo:
Some of the extension services in this repository follows a very bad example for other extension authors by not having files in the container rootfs and instead bind mounting it from the host, this started as a simple copy paste as we were experimenting with extensions and some extensions especially nvidia needs files both on host and for extension services.

Go through each extension service and make each of them self contained unless the files needs to be present on the host also.

PR #962, opened 2026-01-30
feat: consolidate extension services

frezbo:
Consolidate extension services to be self contained, don't pollute rootfs or mount up libraries from host unless absolutely needed.

Fixes: #876
Fixes: https://github.com/siderolabs/talos/issues/12667


.kres.yaml
```diff
@@ -277,6 +280,18 @@ spec:
...
+kind: custom.Step
+name: test-%
+spec:
+  makefile:
+    enabled: true
+    script:
+      - |
+        @$(MAKE) docker-test-extension TARGET_ARGS="--tag=$(REGISTRY)/$(USERNAME)/$@:v0 --push=$(PUSH)" TEST_STAGE=$*
+        @docker pull $(REGISTRY)/$(USERNAME)/$@:v0
+        @orb debug $(REGISTRY)/$(USERNAME)/$@:v0
```
dsseng, line 292: Could there be a way to do this with Docker or another Linux tool?

dsseng: Perhaps `BUILDX_EXPERIMENTAL=1 docker buildx debug build`?

power/nut-client/pkg.yaml
```diff
@@ -21,11 +30,9 @@ steps:
 
         export PKG_CONFIG_PATH=/usr/lib/pkgconfig
         ./configure \
-                --prefix=/usr/local \
-                --libexecdir=/usr/local/lib/nut \
-                --datadir=/usr/local/share/nut \
+                --prefix=/usr/local/lib/containers/nut-client \
+                --libexecdir=/usr/local/lib/containers/nut-client/libexec \
```
dsseng, line 34: Have you checked if this builds correctly? Oftentimes such args expect a relative path under prefix

storage/multipath-tools/pkg.yaml
```diff
@@ -34,25 +60,26 @@ steps:
           SYSTEMD=""
     install:
       - |
-        mkdir -p /rootfs/usr/local/lib/containers/multipathd/ /rootfs/usr/local/lib/
-        cp /usr/lib/libgcc_s.so.1 /rootfs/usr/local/lib/
-        make prefix="/usr/local" DESTDIR=/rootfs LIB=lib install
+        mkdir -p /rootfs/usr/local/lib/containers/multipath-tools/usr/local/lib/
+
+        cp /usr/lib/libgcc_s.so.1 /rootfs/usr/local/lib/containers/multipath-tools/usr/local/lib/
```
dsseng, line 65: Doesn't copying from Musl package add this as well by chance? Not sure, but could be

.kres.yaml
frezbo: it won't since our images don't have shell, orb debug uses a nix environment which is self contained at `/nix` so just set `PATH` to `/nix` and everything works, I wished we could, or hack around with a readymade nix image

storage/multipath-tools/pkg.yaml
frezbo: you meant to copy from musl from stagex? or tools

power/nut-client/pkg.yaml
frezbo: I tried to fix as many packages as possible and put the right prefix, this i believe is for the binaries

storage/multipath-tools/pkg.yaml
dsseng: Ah, I guess it is not a part of musl, nvm

PR #977, opened 2026-02-09
feat: add FRR (Free Range Routing) extension

### LINKED ISSUE #625: consider pulling in FRR extension
See https://github.com/abckey/frr-talos-extension

maxpain:
## Description

Adds FRR (Free Range Routing) 10.5.1 extension for Talos Linux with a per-daemon architecture: each FRR daemon runs as a separate Talos extension service, sharing a common container root via bind mounts.

## Architecture

Each daemon is a standalone extension service with its own dependency chain:

```
ext-frr-mgmtd  (centralized config manager)
  └─ ext-frr-zebra  (kernel routing manager)
       ├─ ext-frr-staticd  (static routes)
       ├─ ext-frr-bgpd  (BGP)
       └─ ext-frr-bfdd  (BFD)
```

All daemons start without configuration and work idle.

### Configuration model

FRR 10.x uses `mgmtd` as the centralized configuration manager. `mgmtd` reads config files for zebra and staticd (`zebra.conf`, `staticd.conf`) and pushes configuration to those daemons via internal protocol. `bgpd` and `bfdd` manage their own config files directly.

Therefore, `ExtensionServiceConfig` documents should target:

| Config for | ExtensionServiceConfig `name` | Mount path |
|------------|-------------------------------|------------|
| zebra (interfaces, RA, addresses) | `frr-mgmtd` | `/usr/local/etc/frr/zebra.conf` |
| staticd (static routes) | `frr-mgmtd` | `/usr/local/etc/frr/staticd.conf` |
| bgpd (BGP) | `frr-bgpd` | `/usr/local/etc/frr/bgpd.conf` |
| bfdd (BFD) | `frr-bfdd` | `/usr/local/etc/frr/bfdd.conf` |

Only the five daemons above are included. Other daemons (ospfd, ripd, isisd, etc.) are disabled at `./configure` time because they are unnecessary for most users. Since Talos currently has no mechanism to disable individual services from a multi-service extension (see https://github.com/siderolabs/talos/issues/12764), every built daemon starts unconditionally. Additional daemons (e.g. ospfd, isisd) can be enabled in future versions when needed.

## Implementation Details

- Per-daemon extension services (no watchfrr/bash/busybox supervisor)
- Shared FRR container root with per-daemon minimal roots (only /etc/passwd, /etc/group, /usr/local/etc/frr/)
- FRR binaries bind-mounted read-only from shared root into each daemon container
- Shared library build (json-c from pkgs, libyang as sub-package)
- Bundles required shared libraries (libyang, libjson-c, libpcre2, libatomic)
- Workaround for upstream FRRouting/frr#15752 (build fails with `--disable-vtysh`)
- vtysh disabled (not needed in Talos containerized environment)
- Pre-created empty `.conf` files as bind-mount targets for `ExtensionServiceConfig`
- Service YAML discovery via `/pkg/frr-*.yaml` glob — adding a new daemon requires only a new YAML file

## Configuration Example

```yaml
# Zebra interface config — mounted into mgmtd container
apiVersion: v1alpha1
kind: ExtensionServiceConfig
name: frr-mgmtd
configFiles:
  - mountPath: /usr/local/etc/frr/zebra.conf
    content: |
      frr defaults datacenter
      !
      interface lo
        ipv6 address 5f00:1:3::/128
---
# BGP config — mounted into bgpd container
apiVersion: v1alpha1
kind: ExtensionServiceConfig
name: frr-bgpd
configFiles:
  - mountPath: /usr/local/etc/frr/bgpd.conf
    content: |
      frr defaults datacenter
      !
      router bgp 65000
        bgp router-id 10.0.0.1
        neighbor FABRIC peer-group
        neighbor FABRIC remote-as external
        neighbor eth0 interface v6only peer-group FABRIC
```

## Related

- Closes #625
- Feature request for optional extension services: https://github.com/siderolabs/talos/issues/12764


Preisschild:
> Is the multi-daemon + supervisor pattern acceptable for this use case?

Not a maintainer but you can also have multiple containers inside one extension fwiw. I have been doing this for tailscale and a route manager for it

Cool PR though!

maxpain:
Thanks! I'm actually reworking this right now to use separate extension services (one per daemon) instead of the watchfrr supervisor approach. Should have an update soon.

One thing I'm running into: FRR has many daemons (zebra, bgpd, ospfd, bfdd, isisd, etc.), but most users only need a few of them. I don't want to register all of them as extension services because they'd all start automatically, even if unused.

Ideally there would be something like an `ExtensionConfig` (as opposed to `ExtensionServiceConfig`) — a single config document for the whole extension where you could declare which services should be enabled. Right now `ExtensionServiceConfig` is per-service, so there's no way to centrally control which daemons to start. cc @smira @dsseng — would something like this be feasible in Talos?

smira:
> Thanks! I'm actually reworking this right now to use separate extension services (one per daemon) instead of the watchfrr supervisor approach. Should have an update soon.
> 
> One thing I'm running into: FRR has many daemons (zebra, bgpd, ospfd, bfdd, isisd, etc.), but most users only need a few of them. I don't want to register all of them as extension services because they'd all start automatically, even if unused.
> 
> Ideally there would be something like an `ExtensionConfig` (as opposed to `ExtensionServiceConfig`) — a single config document for the whole extension where you could declare which services should be enabled. Right now `ExtensionServiceConfig` is per-service, so there's no way to centrally control which daemons to start. cc @smira @dsseng — would something like this be feasible in Talos?

Does this actually need to be a system extension? Can this be run as a static pod/daemonset? I think there is some limit when this becomes a mess that is hard to manage/doesn't make sense to manage at the OS level.

PR #1000, opened 2026-03-09
feat: update dependencies

smira:
Pull in new pkgs, and bump:

* ZFS 2.4.1
* NVIDIA LTS 580.126.20

Bumps in extensions:

* netbird 0.66.2
* newt 1.9.0
* tailscale 1.94.2
* spin 0.23.0
* youki 0.6.0
* Intel ucode 20260227


drivers/vars.yaml
```diff
@@ -1,4 +1,4 @@
 # renovate: datasource=github-releases extractVersion=^v(?<version>.*)$ depName=systemd/systemd
-SYSTEMD_VERSION: 259.1
+SYSTEMD_VERSION: 259.3
 SYSTEMD_SHA256: d5a6c72c7c3a50e251f74c15672a47f63674aa12eda64086a7daffa49c4c55f8
```
frezbo, line 3: sha's not updated?

smira: good question.. let me check

smira: it downloads just a single udevd rule, that's why checksum doesn't change

frezbo (approved):

PR #1011, opened 2026-03-17
fix: rework image digests one more time

smira:
For some reason buildx in the CI doesn't want to export full metadata, only the digest is exposed.

Use a mixed solution by building the image tag via bldr as previously, while attaching the digest from buildx.

### DIFF
--- a/.kres.yaml
+++ b/.kres.yaml
@@ -123,6 +123,8 @@ spec:
       toplevel: true
     - name: check-dirty
       toplevel: true
+    - name: extensions-metadata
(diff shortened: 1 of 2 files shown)


.kres.yaml
```diff
@@ -183,6 +185,18 @@ spec:
     condition: on-pull-request
 ---
 kind: custom.Step
+name: extensions-metadata
```
smira, line 188: this is for backwards compat with `integration-extensions` in Talos

PR #1028, opened 2026-03-29
Export zpools on zfs shutdown

gjabell:
Fixes cases where later tasks fail because the disks are still in use (e.g. closing encrypted volumes).

I ran into similar issues as https://github.com/siderolabs/talos/issues/9307 and https://github.com/siderolabs/talos/issues/8800 while setting up a new cluster with a ZFS data disk; since the pools are still active, `cryptsetup close` fails in a loop until it reaches the timeout and this also prevents the normal upgrade process from completing (`--stage` can still be used to upgrade but it takes much longer than necessary).

### DIFF
--- a/storage/zfs/zfs-service/main.go
+++ b/storage/zfs/zfs-service/main.go
@@ -31,4 +31,11 @@ func main() {
 	if err := cmd.Run(); err != nil {
 		log.Fatalf("zfs-service: zfs unmount error: %v\n", err)
 	}
+
+	cmd = exec.Command("/usr/local/sbin/zpool", "export", "-a")
+	cmd.Stdout = os.Stdout
+	cmd.Stderr = os.Stderr
+	if err := cmd.Run(); err != nil {
+		log.Fatalf("zfs-service: zpool export error: %v\n", err)
+	}
 }


storage/zfs/zfs-service/main.go
```diff
@@ -31,4 +31,11 @@ func main() {
 	if err := cmd.Run(); err != nil {
 		log.Fatalf("zfs-service: zfs unmount error: %v\n", err)
 	}
+
+	cmd = exec.Command("/usr/local/sbin/zpool", "export", "-a")
```
frezbo, line 35: we do an umount before, isn't that not enough?

gjabell: From what I understand, until the zpool is exported, the block devices are still considered to be in use (hence why running `cryptsetup close` will fail): https://github.com/openzfs/zfs/issues/17773

frezbo: From:

```
To my understanding, the pool's mounts are being unmounted as part of the unmounting process during shutdown which, and that blocks cryptsetup from closing the associated LUKS volumes
```

So i guess the problem is `cryptsetup close` called after zfs unmount?

gjabell: Yes; not sure whether there are other cases where not exporting zfs would be an issue as that's the only one I've run into. I think either skipping the `cryptsetup close` or exporting the zfs pool would be a reasonable solution, but I assumed the cryptsetup hook is more general and therefore harder to change.

frezbo: yes, export seems easier than hooking into when `cryptsetup close` gets called

PR #1048, opened 2026-04-16
chore: ci test for relative symlinks

frezbo:
CI tests for relative symlinks.


smira:
As discussed with @frezbo, this can be turned out into a validation and a fix to drop relative symlinks.

smira:
Confirm whether `go-containerregistry` now allows relative symlinks.

PR #1089, opened 2026-05-20
fix: add zlib & zstd to btrfsprogs

smira:
This is required to run btrfsprogs.


smira:
I need to see if I can rework this to either use static or move libs to Talos rootfs.

PR #1121, opened 2026-06-18
chore(drm/xe): ship i915 firmware directory

ThePhaseless:
The "xe" driver needs i915 DMC firmware (e.g. "i915/bmg_dmc.bin") to enable runtime power management on Intel Arc/Battlemage GPUs. Copy the full i915 firmware directory into the xe extension rootfs so the driver can find it without requiring the separate i915 extension.


ThePhaseless:
Closing as requested - keeping the fork/branch available for a future PR.

PR #1126, opened 2026-06-19
feat: add hyperv-guest-agent extension

rmvangun:
Adds a Hyper-V guest integration extension under `guest-agents/`. Talos already ships the kernel side (`hv_vmbus` built in, `hv_utils` as a module) but nothing runs the user-space daemons, so the Hyper-V host can't see the guest IP and checkpoints come out crash-consistent only. This came up in siderolabs/talos#6435 and discussion https://github.com/siderolabs/talos/discussions/12721.

It packages the in-tree Linux `tools/hv` daemons as two extension services:

- `ext-hyperv-kvp` (`hv_kvp_daemon`) reports the guest IP, hostname and OS to the host, so `Get-VMNetworkAdapter` gets populated.
- `ext-hyperv-vss` (`hv_vss_daemon`) freezes the `/var` data partition for application-consistent checkpoints.

A few notes on the build:

- Compiled statically from the kernel tree at a pinned, checksummed tag, tracked by Renovate in `guest-agents/vars.yaml`. Same approach as `qemu-guest-agent`/`xen-guest-agent`.
- VSS uses the same host-mount pattern as the `zfs` extension (`rootfsPropagation: shared`) so the `FIFREEZE` lands on the real data partition.
- Tier `extra`; I've added myself to MAINTAINERS.md.

One thing documented in the README: KVP's DNS/DHCP reporting and host-to-guest IP injection rely on the distro helper scripts (`hv_get_dns_info` and friends) that this extension doesn't ship, so those stay empty. IP/hostname/OS reporting doesn't need them.


rmvangun:
Confirmed working on real Hyper-V — fully validated on a Talos node.

smira (commented):
lgtm, @frezbo your thoughts?

guest-agents/hyperv-guest-agent/hyperv-kvp.yaml
```diff
@@ -0,0 +1,46 @@
...
+      destination: /sys
+      type: bind
+      options:
+        - rbind
+        - ro
+    # Writeable location for the KVP pool files (.kvp_pool_0..3). The daemon's
+    # compiled-in KVP_CONFIG_LOC is /var/lib/hyperv; it tolerates the directory
+    # already existing, so a writeable bind-mount satisfies it without patching
+    # the upstream source.
+    - source: /system/run/hyperv-kvp
```
frezbo: let's just use /run

guest-agents/hyperv-guest-agent/hyperv-vss.yaml
```diff
@@ -0,0 +1,43 @@
...
+        - rshared
+        - rbind
+        - rw
+    - source: /run
+      destination: /run
+      type: bind
+      options:
+        - rshared
+        - rbind
+        - rw
```
frezbo: i guess this has some socket?

guest-agents/vars.yaml
```diff
@@ -18,3 +18,12 @@ XEN_GUEST_AGENT_SHA512: 49bf15d7257f7fcb5ac919ca57e8c16bb6f8199684adef034bd1e768
 TALOS_VMTOOLSD_VERSION: v1.5.0
 # renovate: datasource=github-releases depName=siderolabs/talos-metal-agent
 TALOS_METAL_AGENT_VERSION: v0.1.5
+# hyperv-guest-agent builds hv_kvp_daemon and hv_vss_daemon from the in-tree kernel
+# tools/hv sources. gregkh/linux is the official linux-stable mirror; pin to a stable
+# patch tag (the tools/hv interface is stable across patch releases).
+# renovate: datasource=git-tags extractVersion=^v(?<version>.*)$ depName=https://github.com/gregkh/linux.git
+LINUX_HV_VERSION: 6.12.28
```
frezbo: why 6.12 instead of 6.18?

guest-agents/hyperv-guest-agent/pkg.yaml
```diff
@@ -0,0 +1,57 @@
...
+steps:
+  # hv_kvp_daemon and hv_vss_daemon are the upstream Linux Hyper-V integration
+  # daemons that live in the kernel tree under tools/hv. They are plain C with no
+  # dependencies beyond libc, so they are fetched (checksum-verified) and linked
+  # statically against the base toolchain — the same model qemu-guest-agent uses
+  # for a static guest agent. gregkh/linux is the official linux-stable GitHub
+  # mirror; only patch tags (e.g. v6.12.28) live there, and the tools/hv interface
+  # is stable across patch releases.
+  - sources:
+      - url: https://raw.githubusercontent.com/gregkh/linux/v{{ .LINUX_HV_VERSION }}/tools/hv/hv_kvp_daemon.c
```
frezbo: let's just use the kernel release tarball and use that, it's more stable

frezbo (approved):
🆒

guest-agents/hyperv-guest-agent/hyperv-kvp.yaml
rmvangun: OK

guest-agents/hyperv-guest-agent/hyperv-vss.yaml
rmvangun: No, remove the /var mount.

guest-agents/hyperv-guest-agent/pkg.yaml
rmvangun: OK

guest-agents/vars.yaml
rmvangun: Moved to 6.18

rmvangun:
Addressed those comments and tested it again on a hyperv workbench.

guest-agents/vars.yaml
```diff
@@ -18,3 +18,9 @@ XEN_GUEST_AGENT_SHA512: 49bf15d7257f7fcb5ac919ca57e8c16bb6f8199684adef034bd1e768
 TALOS_VMTOOLSD_VERSION: v1.5.0
 # renovate: datasource=github-releases depName=siderolabs/talos-metal-agent
 TALOS_METAL_AGENT_VERSION: v0.1.5
+# hyperv-guest-agent builds hv_kvp_daemon and hv_vss_daemon from the in-tree
+# kernel tools/hv sources extracted from the official kernel.org release tarball.
+# renovate: datasource=git-tags extractVersion=^v(?<version>.*)$ depName=https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git
```
frezbo: notes to self:

before merge use the same rule for kernel we use in pkgs and add the renovate lock, so we don't get anything above 6.18

frezbo (approved):

frezbo (approved):

PR #1127, opened 2026-06-21
docs: document how to build a single extensions and a catalog

Orzelius:
Modify the makefile a bit so extensions and a catalog can be built and pushed locally. This will be useful once people are making their own extra extensions.


.kres.yaml
```diff
@@ -227,9 +229,9 @@ spec:
     script:
       - |
         @echo "Generating image descriptions..."
-        @echo -n "" > internal/extensions/descriptions.yaml
+        @printf '' > internal/extensions/descriptions.yaml
```
Orzelius, line 230: -n doesn't work on macos

README.md
```diff
@@ -213,6 +213,47 @@ The rough flow should look like the following:
 - You can now build your extension using make like `make <extension-name> PLATFORM=linux/amd64`
 - If you wish to output the contents of the image and validate your install, you can issue `make local-<extension-name> PLATFORM=linux/amd64 DEST=_out`. The contents will then be present in the `_out` directory.
 
+### Pushing Images Locally
+
+To build and push an extension to a local registry use:
+
+```bash
+make <extension-name> PLATFORM=linux/arm64 PUSH=true REGISTRY=host.docker.internal:5005 USERNAME=<your-name>
```
smira: I think we should say to create the buildkit runner with host networkinghttps://docs.siderolabs.com/talos/v1.13/build-and-extend-talos/custom-images-and-development/developing-talos#prepare

This drops all the `--insecure` problems and simplifies the flow down to using `127.0.0.1:5005`

If you want to be complete here, I would add a note on running a local registry

This kills CRANE_FLAGS and all other issues. 

(--insecure is automatically applied for localhost IPs)

Orzelius: thank you! I don't know why I didn't think of this, makes perfect sense

.kres.yaml
smira: I think we expect people to install Linux coreutils via `brew`

smira (approved):

PR #1137, opened 2026-06-29
feat: add talosctl-oidc extension (contrib)

qjoly:
This PR adds `talosctl-oidc` as a contrib system extension under `misc/`.

`talosctl-oidc` is an OIDC certificate exchange server for Talos Linux. It validates OIDC tokens and issues short-lived client certificates signed by the Talos CA, so `talosctl` access can be controlled through OIDC. Talos has no native OIDC support, so the server covers that case.

The extension runs the server as a Talos extension service. The binary comes from the project's GitHub releases (linux amd64 and arm64), with sha256 and sha512 pinned in `misc/vars.yaml`.

Tier: contrib.

One thing to flag: the server needs the Talos CA private key, which a read-only extension rootfs cannot carry. It has to be supplied at runtime through an `ExtensionServiceConfig` or a mounted volume.

Project: https://github.com/qjoly/talosctl-oidc


misc/talosctl-oidc/pkg.yaml
```diff
@@ -0,0 +1,46 @@
...
+  - stage: base
+steps:
+  - sources:
+      # {{ if eq .ARCH "aarch64" }} YAML comment; Go template evaluated by bldr
+      - url: https://github.com/qjoly/talosctl-oidc/releases/download/{{ .TALOSCTL_OIDC_VERSION }}/talosctl-oidc-linux-arm64
+        destination: talosctl-oidc
+        sha256: {{ .TALOSCTL_OIDC_ARM64_SHA256 }}
+        sha512: {{ .TALOSCTL_OIDC_ARM64_SHA512 }}
+      # {{ else if eq .ARCH "x86_64" }} YAML comment; Go template evaluated by bldr
+      - url: https://github.com/qjoly/talosctl-oidc/releases/download/{{ .TALOSCTL_OIDC_VERSION }}/talosctl-oidc-linux-amd64
```
smira, line 14: this should be built from source.

also not sure why it should be an extension vs. e.g. a static pod? how does it integrate with Talos API?

qjoly: > this should be built from source.

I update my PR, thanks 👍 

> also not sure why it should be an extension vs. e.g. a static pod? how does it integrate with Talos API?

I support both methods:
 - Pod: less secure, since we have to give it the CA private key, which any cluster operator can read.
 - Extension: less user-friendly, but more secure, since only Talos administrators can read the key.

The software doesn't need to integrate with the Talos API. It only generates short-lived certificates and can run anywhere, without communicating with any Talos node.

smira: Why should it be given any CA? 

It can run perfectly in Kubernetes, use Talos API access via Kubernetes, and call the API to generate talosconfig. No need to expose CA.

qjoly: > Talos API access via Kubernetes

To generate short-lived certificates, talosctl-oidc needs the Talos CA (certificate and private key) mounted into it. The built-in "Talos API access through Kubernetes" feature isn't enough here: it only hands out a single fixed pair of certificates and can't mint new ones per user.

smira: It can via Talos API, Talos API has a way to generate new client certificates - it's really bad idea to give it cluster CA.

qjoly: You were right, I can absolutely generate new client certificates through Talos API access via Kubernetes.

I'm closing this PR and deprecating the Talos extension for now 🙂 I'll likely bring it back later by auto-discovering the admin certificate (e.g. by mounting the virtual device that holds the config).

smira: You mean Talos certificate? 

You need to do equivalent of `talosctl config new` via Talos API to get `talosconfig` - you control the roles & TTL. 

So you just need to accept incoming JWT token, validate it, and return in exchange new `talosconfig`. So you can make it a Kubernetes service, or NodePort or whatever else makes sense.

PR #1138, opened 2026-07-01
chore: make mdadm package no-op

shanduur:
Move mdadm from extensions into core packages.


smira (approved):

frezbo (approved):
I wonder if we can do something in factory to map deprecated extensions and drop all this here

smira:
> I wonder if we can do something in factory to map deprecated extensions and drop all this here

we could, but it might confuse Omni or other automation, as the schematic installs an extension, while the actual Talos image doesn't have it.

we could replace it with empty extension in the Factory though

PR #1141, opened 2026-07-03
feat: add trident-nfs-tools extension for NFS support on Talos

ohauer:
## Summary

Adds a new `trident-nfs-tools` system extension that provides `/usr/local/sbin/mkdir` (busybox) on the Talos host.

## Problem

Trident's `chwrap` mechanism shells out to `mkdir` to create NFS mountpoint directories before calling `mount.nfs4`. Talos Linux does not ship `mkdir` or any coreutils, causing all NFS mounts to fail:

```
Mkdir failed: exit status 2
error mounting NFS volume ... exit status 32
```

The iSCSI code path uses Go's native `os.MkdirAll()` and is not affected. Only the NFS path shells out to the binary.

## What's Included

- `/usr/local/sbin/mkdir` — busybox multi-call binary (only depends on musl libc)

## Testing

- Confirmed `mkdir` is not present on Talos v1.13.x hosts
- Confirmed chwrap searches `/usr/local/sbin/` and will find the binary
- Extension passes `extensions-validator`

## Related

- Trident issue: https://github.com/NetApp/trident/issues/806
- Trident upstream fix (pending): https://github.com/NetApp/trident/pull/1152
- Pattern follows existing `trident-iscsi-tools` extension

## Note

Once Trident PR #1152 is merged (replaces `mkdir` shell-out with `os.MkdirAll()`), this extension will no longer be needed. Until then, it unblocks NFS on Talos.


frezbo:
If there's a PR upstream i would suggest to wait until it's available, it;s not simple to remove an extension once it's published, there's a lot of downstream consumers that would need extra code to handle unpublished extensions

smira:
@kevintijssen your opinion?

kevintijssen:
> @kevintijssen your opinion?

I agree with @frezbo. I have tested NFSv4 without any issues so far. I noticed the `mkdir` message when using nfs but did not had any impact.

smira:
> > @kevintijssen your opinion?
> 
> I agree with @frezbo. I have tested NFSv4 without any issues so far. I noticed the `mkdir` message when using nfs but did not had any impact.

thanks, I put it on hold for now

ohauer:
I can confirm trident PVC can created without mkdir util, it is only a cosmetic issue and i agree with close/reject the request

PR #1155, opened 2026-07-17
feat: add kata-containers-snp extension

calintkd:
Adds `container-runtime/kata-containers-snp`: AMD SEV-SNP confidential VMs
for Kata Containers ([Confidential Containers](https://confidentialcontainers.org/)),
registering the `kata-qemu-snp` and `kata-qemu-coco-dev` containerd runtime
handlers as a layer on top of the in-catalog `kata-containers` extension —
continuing the SEV-SNP enablement from siderolabs/pkgs#1396 (host SNP
kernel) and siderolabs/pkgs#1608 (IOMMUFD).

### What it ships (amd64 only)

From the same sha256-pinned `kata-static` release tarball the
`kata-containers` extension already uses, only the confidential delta
(~92 MB): the `AMDSEV.fd` OVMF (the SNP-capable firmware), the confidential
guest kernel + dm-verity guest image, and the two handler configs. The
shim, `virtiofsd`, and the QEMU these handlers run on come from the
`kata-containers` extension (required on the same node; the README and
description state it). The two extensions share no files, and their
versions stay paired through the catalog's single `KATA_CONTAINERS_VERSION`
pin.

The configs are generated at build time from the release's own defaults
(same approach as #1157) with three edits: `/opt/kata` → `/usr/local` path
rewrite, `experimental_force_guest_pull = true` (Talos cannot run the nydus
snapshotter host daemon kata-deploy defaults to; the config flag is the
deployable path), and `create_container_timeout` 60 → 1200 s (guest-pull
unpacks multi-GB images inside the guest during CreateContainerRequest).
Everything else — including the guest image's dm-verity parameters — comes
from the release itself, so a Kata version bump needs no config maintenance
here. The test phase asserts the exact final state of every generated field
(hypervisor path, guest-pull, timeout, verity params present, no `/opt/kata`
remnants), so an upstream rename or format change fails the build instead of
silently dropping an edit.

On arm64 the extension builds as an empty rootfs + manifest (the SNP payload
is amd64-only), following the `drivers/mei` / `drm/intel-npu` pattern.

### Why a layer on top of `kata-containers`

SEV-SNP cannot be added to the existing extension's default runtime: Kata's
Cloud Hypervisor driver rejects it (`SEV-SNP protection is not supported by
Cloud Hypervisor`, `virtcontainers/clh.go`) — SNP in Kata is QEMU-only.
With #1157, `kata-containers` ships the standard QEMU + datadir these
handlers run on, so this extension no longer needs to carry a hypervisor at
all: it adds only the confidential guest stack, and only for users who want
it. No duplicated files means the two extensions co-install cleanly (no
`exceptions-amd64.yaml` entries needed) and can be tested together in the
same e2e install. The extension is also registered in the internal grype
scan set (`internal/grype-scan/pkg.yaml`).

### Testing

- `kata-qemu-snp` validated on bare-metal AMD EPYC (Turin) with SEV-SNP:
  guest boots with `Memory Encryption Features active: AMD SEV SEV-ES SEV-SNP`,
  attestation reports issued by the PSP inside the guest, SNP launch
  measurement verified against a computed reference value.
- The generated configs are byte-identical (modulo the release's own
  auto-generated header comment) to the configs those runs validated, and
  the OVMF/guest kernel/guest image are the same tarball objects — the
  measured launch path is unchanged by the layering.
- `kata-qemu-coco-dev` exercises the same guest stack on plain KVM — this is
  CI-testable without SNP hardware; with no file overlap against
  `kata-containers`, the QEMU e2e lane can install both extensions together
  and drive a `testRuntimeClass("kata-qemu-coco-dev", ...)` case in
  `internal/integration/api/extensions_qemu.go` in siderolabs/talos — happy
  to submit that as a follow-up, same as the existing `kata` test.
- `>= v1.12.0` floor: host-side SEV-SNP kernel support entered Talos in v1.12
  (pkgs#1396); validated on v1.13.5.

### Open questions for review

- Tier: proposed `extra` to match `kata-containers`.
- `kata-qemu-coco-dev` is included because it is the CI-exercisable handler
  (same guest stack, no TEE hardware needed) — it is what a
  `testRuntimeClass` case in the Talos QEMU lane would drive. Can drop it to
  SNP-only if you prefer a single-purpose extension.


container-runtime/kata-containers-snp/configuration-qemu-coco-dev.toml
```diff
@@ -0,0 +1,769 @@
...
+# Do not set values here unless you understand the impact of doing so as you
+# may stop the virtual machine from booting.
+# To see the list of default parameters, enable hypervisor debug, create a
+# container and look for 'default-kernel-parameters' log entries.
+kernel_params = ""
+
+# Optional dm-verity parameters (comma-separated key=value list):
+# root_hash=...,salt=...,data_blocks=...,data_block_size=...,hash_block_size=...
+# These are used by the runtime to assemble dm-verity kernel params.
+kernel_verity_params = "root_hash=c1b7e41bcb256b64df44bd5bccd1d05517e581136719caaa7731cbec2eebdde3,salt=4ce0ac520506b38fb03281ccbc18300d439007c67d6f5ed7cbf23b8a692640c5,data_blocks=96512,data_block_size=4096,hash_block_size=4096"
```
smira: would updating Kata version invalidate those?

calintkd: Yes — the confidential guest image is rebuilt for each release, so its `root_hash`/`salt`/`data_blocks` change with it. That's what the build gate in the `test:` phase is for: it diffs the vendored `kernel_verity_params` against the release tarball's own defaults (both configs), so bumping `KATA_CONTAINERS_VERSION` without re-vendoring fails the build instead of shipping a guest that fails dm-verity verification at boot.

The configs are vendored so the final result stays reviewable — they're the release's own defaults with only the three edits named in the description (`/opt/kata` → `/usr/local`, guest-pull, `create_container_timeout`), verity params carried verbatim; the bump remedy is re-vendoring from the new release's defaults and re-applying those edits.

smira: I wonder if it's better to do approach like https://github.com/siderolabs/extensions/pull/1157 which does `sed` on the fields which need to be adjusted? I would like to get maintenance on upgrades to be zero. 

Plus right now as this extension conflicts with kata-containers, we would probably not have a test coverage for it.

calintkd: Done — switched the configs to generation from the release's own defaults (same approach as #1157): path rewrite, guest-pull, `create_container_timeout`; everything else, including the dm-verity parameters, comes from the release itself, so a Kata bump needs no maintenance here. Each generated field is asserted after generation, so an upstream rename fails the build rather than silently dropping an edit.

And you're right about the conflict — #1157 removed the reason this extension carried a hypervisor at all. It's now a pure SNP layer on top of `kata-containers`: dropped the shim/virtiofsd/QEMU/datadir entirely; it ships only `AMDSEV.fd`, the confidential guest kernel + dm-verity image, and the two handler configs (~92 MB, down from ~150). No shared files, so they co-install cleanly and the exceptions entries are gone — which should also unblock e2e coverage; happy to add the `testRuntimeClass("kata-qemu-coco-dev")` case as a follow-up. The generated configs come out byte-identical to the previously vendored (hardware-validated) ones, so the tested launch path is unchanged.

smira (approved):
thank you!

PR #1157, opened 2026-07-19
feat(kata-containers): add QEMU hypervisor handler

onedr0p:
Hi 👋 

This adds a `kata-qemu` containerd runtime handler alongside the existing Cloud Hypervisor one, selectable per-workload with a `RuntimeClass`.

**Why:** I run GitHub Actions runners on-cluster, and some CI jobs boot full VMs with QEMU (e.g. e2e tests that spin up a Talos cluster). Today that forces the runner pod to be fully privileged on the host kernel. The goal is to run those jobs inside a Kata VM instead, so the privilege stays behind a real VM boundary - but that requires nested virtualization inside the Kata guest. Cloud Hypervisor does not expose the CPU virtualization flag to the guest, so the existing `kata` handler can't do this; QEMU with `-cpu host` does (on x86-64).

The workflow boots a 3-node Talos cluster with the QEMU provisioner:

```bash
sudo -E "$(mise which talosctl)" cluster create qemu \
  --name "$cluster" \
  --cidr 10.5.0.0/24 \
  --schematic-id "$schematic" \
  --controlplanes 1 \
  --workers 2 \
  --memory-workers 5GiB \
  --disks "virtio:8GiB,virtio:${MIROIR_E2E_DISK_SIZE:-20GiB}" \
  --config-patch @deploy/e2e/talos-patch.yaml \
  --talosconfig-destination "$TALOSCONFIG"
```

**Size impact (x86-64):** I know there is a limitation so I wanted to give some data on the size difference this would be adding. Stripped `qemu-system-x86_64` (~30MB) plus its pruned datadir (<1MB after removing firmware for platforms kata never emulates); measured on the assembled extension rootfs:

| | uncompressed | squashfs (zstd) |
| --- | --- | --- |
| before | ~386MB | ~74MiB |
| after | ~416MB | ~78.7MiB |

**Question:** would you prefer this as a separate system extension (e.g. `kata-containers-qemu`) instead of bundling QEMU into `kata-containers`? Happy to split it out.

### DIFF
--- a/container-runtime/kata-containers/pkg.yaml
+++ b/container-runtime/kata-containers/pkg.yaml
@@ -61,6 +61,40 @@ steps:
 
         mkdir -p /rootfs/usr/local/share/kata-containers
         cp /pkg/configuration.toml /rootfs/usr/local/share/kata-containers/
+      - |
+        cd kata-static
+
+        qemu={{ if eq .ARCH "aarch64" }}qemu-system-aarch64{{ else }}qemu-system-x86_64{{ end }}
+        cp ./opt/kata/bin/${qemu} /rootfs/usr/local/bin/${qemu}
+        strip /rootfs/usr/local/bin/${qemu}
+
+        mkdir -p /rootfs/usr/local/share/kata-qemu
+        cp -a ./opt/kata/share/kata-qemu/qemu /rootfs/usr/local/share/kata-qemu/qemu
+        rm -f /rootfs/usr/local/share/kata-qemu/qemu/edk2-*.fd
+        rm -rf /rootfs/usr/local/share/kata-qemu/qemu/firmware
+        # ROMs for platforms kata's q35/virt machines never use (hppa, ppc, s390, arm boards)
+        rm -f /rootfs/usr/local/share/kata-qemu/qemu/{hppa-firmware*.img,s390-*.img,slof.bin,vof*.bin,qemu_vga.ndrv,pnv-pnor.bin,*_bootrom.bin,qemu-nsis.bmp}
+
+        # kata builds QEMU with --disable-relocatable, so the firmware search path is
+        # hardcoded to /opt/kata/share/kata-qemu/qemu. QEMU still consults
+        # <exec_dir>/qemu-bundle/<compiled path> first; route that back to the datadir
+        mkdir -p /rootfs/usr/local/bin/qemu-bundle/opt/kata/share/kata-qemu
+        ln -s ../../../../../../share/kata-qemu/qemu /rootfs/usr/local/bin/qemu-bundle/opt/kata/share/kata-qemu/qemu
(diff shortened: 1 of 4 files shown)


smira (approved):
note: follow up with a test on Talos side

container-runtime/kata-containers/pkg.yaml
```diff
@@ -61,6 +61,34 @@ steps:
...
+        rm -f /rootfs/usr/local/share/kata-qemu/qemu/{hppa-firmware*.img,s390-*.img,slof.bin,vof*.bin,qemu_vga.ndrv,pnv-pnor.bin,*_bootrom.bin,qemu-nsis.bmp}
+        # {{ if eq .ARCH "aarch64" }} This in fact is YAML comment, but Go templating instruction is evaluated by bldr
+        mkdir -p /rootfs/usr/local/share/aavmf
+        cp ./opt/kata/share/aavmf/AAVMF_CODE.fd /rootfs/usr/local/share/aavmf/AAVMF_CODE.fd
+        # {{ end }} This in fact is YAML comment, but Go templating instruction is evaluated by bldr
+
+        # AAVMF_VARS.fd is not shipped: kata ignores firmware_volume for
+        # non-confidential guests, so blank it instead of pointing at a missing file
+        mkdir -p /rootfs/usr/local/share/kata-containers
+        sed -e 's|/opt/kata|/usr/local|g' \
```
smira, line 91: this seems to be broken:

```
172.20.0.6: time="2026-07-21T18:19:55.508962740Z" level=error msg="qemu: could not load PC BIOS 'bios-256k.bin'" name=containerd-shim-v2 pid=1151 qemuPid=1161 sandbox=43689d22536ef0a5f9651ec7842b3598f8f281b1d01dd9af7b7e420198198d5b source=virtcontainers/hypervisor subsystem=qemu
```

smira: seems that Kata is broken with custom locations https://github.com/kata-containers/runtime/issues/2459

smira: firmware_path seems be not working, and firmware doesn't help either

onedr0p: That's an old, archived repo and might be out of date but I believe I have a fix incoming. One moment.

onedr0p: @smira I built this successfully on my local machine with my recent changes. It's a weird one for sure.

Validation:

- `make target-kata-containers PLATFORM=linux/amd64 TARGET_ARGS=--no-cache`
- Clean build completed successfully
- `extensions-validator` passed (`kata-containers:test-2`)

smira: yep, confirmed that it works now, thanks!

Issue #725, opened 2025-06-04
Add support for GPU passthrough with Talos as host (NOT guest)


solidDoWant:
There's a couple of projects that allow using k8s nodes as hypervisors/VM hosts:
* [Kata Containers](https://katacontainers.io/) (supported by Talos via extension)
* [Kubevirt](https://kubevirt.io/) (install documented by Talos [here](https://www.talos.dev/v1.10/advanced/install-kubevirt/))

However neither Talos nor its extensions seem to support GPU passthrough to VMs managed by these tools. Talos is missing some kernel config as documented [here](https://github.com/kata-containers/kata-containers/blob/main/docs/use-cases/Intel-GPU-passthrough-and-Kata.md#host-kernel-requirements), and, in the case of Kata containers, the VM image shipped with the Kata Containers extension is missing some config as defined [here](https://github.com/kata-containers/kata-containers/blob/main/docs/use-cases/Intel-GPU-passthrough-and-Kata.md#install-and-configure-kata-containers) and [here](https://github.com/kata-containers/kata-containers/blob/main/docs/use-cases/Intel-GPU-passthrough-and-Kata.md#build-kata-containers-kernel-with-gpu-support). The Kata Containers project has pretty decent documentation on host + Kata changes required for various hardware accelerator types.

I'm not entirely sure if this should be in https://github.com/siderolabs/extensions or https://github.com/siderolabs/talos (as it involves both the Talos kernel and extensions). Let me know if this issue is in the wrong repo.

smira:
Talos does support device rebind to VFIO, https://www.talos.dev/v1.10/reference/configuration/hardware/pcidriverrebindconfig/

As it comes to specific Kata Containers as a system extension, this is community-supported, so you can take necessary steps for it.

PR #1164, opened 2026-07-23
feat: add kata-containers-nvidia-gpu-snp extension

calintkd:
Adds `container-runtime/kata-containers-nvidia-gpu-snp`: NVIDIA GPU VFIO
passthrough into AMD SEV-SNP confidential VMs for Kata Containers,
registering the `kata-qemu-nvidia-gpu-snp` containerd runtime handler. The
NVIDIA driver and NVRC (NVIDIA's init) run inside the encrypted guest — the
host never loads a GPU driver. Completes the confidential stack on top of
#1155: `kata-containers` (hypervisor base) → `kata-containers-snp` (SNP
layer) → this add-on (confidential GPU) — and answers the ask in #725
(GPU passthrough to Kata VMs).

### What it ships (amd64 only)

From the same sha256-pinned `kata-static` release tarball as its siblings:

- `qemu-system-x86_64-snp-experimental` and its datadir, stripped and
  pruned the same way `kata-containers` treats its QEMU — the GPU+SNP
  config is the only handler that names the snp-experimental binary, so it
  ships here, not in a sibling. No new source or trust root: the binary
  and datadir come from the same sha256-pinned tarball as everything else
  here (kata builds it from the confidential-containers QEMU tree and
  wires only this handler to it). Same `qemu-bundle` datadir redirect
  layout as the `kata-containers` extension.
- the NVIDIA confidential guest kernel and guest image (~533 MiB — one of
  the largest objects in the catalog; the image embeds the NVIDIA driver
  stack and NVRC, which is what keeps the driver inside the TEE boundary).
- the handler config, generated at build time from the release's own
  default with the same recipe as `kata-containers-snp` (`/opt/kata` →
  `/usr/local` rewrite, guest-pull; upstream's NVIDIA default already
  ships `create_container_timeout = 1200`). The dm-verity parameters come
  from the release itself; every generated field is asserted in the test
  phase, and the config's kernel/image paths are asserted to name files
  this extension ships. A Kata bump touches `container-runtime/vars.yaml`
  only (the shared renovate-tracked pin + tarball hashes) — this pkg.yaml
  adds no version-bearing filenames or pins of its own.
- the NVIDIA Driver License Agreement at
  `/usr/local/share/doc/NVIDIA-driver/LICENSE` — the driver license
  permits redistribution provided the agreement text accompanies the
  binaries, and the guest image itself ships none (the same mechanism the
  nvidia-container-toolkit extension applies to its host driver stack,
  `NVIDIA_GLX-1.0/LICENSE`).

The extension ships no CDI writer: the host CDI spec (kind
`nvidia.com/pgpu`, IOMMUFD cdev device nodes) comes from the NVIDIA GPU
Operator's Kata sandbox device plugin, per the flow kata's GPU-passthrough
doc describes — the README documents that setup (`vfioManager.enabled=false`,
since `PCIDriverRebindConfig` already does the binding).

No files are shared with `kata-containers` or `kata-containers-snp` (no
`exceptions-amd64.yaml` entries). On arm64 the extension builds as an empty
rootfs + manifest, as kata-containers-snp does.

### Licensing / SBOM

The sbom declares Apache-2.0 (kata-agent, NVRC), GPL-2.0 (QEMU and its
datadir blobs, guest kernel, guest image userspace), LGPL-2.1 (libseccomp
in the kata-agent), and `CUSTOM` for the NVIDIA
proprietary userspace + GSP firmware inside the guest image — the string
the nonfree-kmod-nvidia sbom in siderolabs/pkgs uses for the driver. The guest kernel modules are built from NVIDIA's open
kernel-module tree; the proprietary part is the redistributable userspace,
the same class of content the nvidia-container-toolkit extension ships.
Happy to switch `CUSTOM` to a different marker if you prefer one.

### Compatibility floor

The real requirement is Talos >= v1.13.7: confidential GPU passthrough
needs `CONFIG_IOMMUFD` (`iommufd.ko`) and `CONFIG_VFIO_DEVICE_CDEV=y` in
the host kernel, which landed on main via siderolabs/pkgs#1608 and
siderolabs/talos#13765 and shipped in the v1.13.7 release (this section
originally said v1.14; corrected 2026-08-05 after the v1.13.7 backport).
The manifest declares `>= v1.13.0`, the highest floor the pinned
`extensions-validator` accepts. The real kernel requirement is v1.13.7
(`CONFIG_IOMMUFD` + `CONFIG_VFIO_DEVICE_CDEV=y`), documented in the README.

### Testing

Validated end-to-end on bare-metal AMD EPYC (Turin) + H200 with GPU CC
mode on: vfio-bound GPU passed into an SEV-SNP guest via the IOMMUFD cdev,
`nvidia-smi` green inside the confidential guest, and the SNP launch
measurement verified against a computed reference value. The generated
config is byte-identical to the config those runs validated, and the host
CDI spec those runs consumed was written to the same shape the operator's
plugin generates on IOMMUFD hosts (kind `nvidia.com/pgpu`, cdev-only
device nodes) — the plugin itself was not part of the rig.
GPU CC mode being off is called out in the README as the classic
silent-hang cause. On a node without a vfio-bound NVIDIA GPU the
extension is inert (the handler is unused), and since no files are shared
with the siblings it does not affect CI coverage of their handlers.

### Open questions for review

- Size: the guest image makes this one of the largest extensions in the
  catalog (~533 MiB). If that is a concern, options are limited — the image is the
  TEE boundary (driver inside the guest), so it cannot be thinned by
  moving the driver to the host.
- Tier: proposed `extra` to match its siblings.
- `CUSTOM` license marker vs something like `LicenseRef-NVIDIA-Proprietary`
  in the sbom, and vendored license text vs fetching it at build time —
  happy to go either way.


frezbo:
>  the host never loads a GPU driver. 

how is this accomplished, how was this tested, did you drop loading nvidia modules from machine config?

calintkd:
> how is this accomplished, how was this tested, did you drop loading nvidia modules from machine config?

On Talos, host NVIDIA kernel modules only come from the NVIDIA kmod extensions (`nvidia-open-gpu-kernel-modules-*` / `nonfree-kmod-nvidia-*`) — this stack installs none of them, so there were no NVIDIA modules on the host to drop; the machine config never loaded one (and the Talos kernel ships no nouveau, so nothing else can claim the device).

The GPU is bound to vfio-pci at boot via the machine config:

```yaml
machine:
  kernel:
    modules:
      - name: vfio_pci
        parameters:
          - ids=10de:233b # H200
```

(`PCIDriverRebindConfig` is the declarative alternative; the README shows both.) The driver stack lives inside the confidential guest image: NVRC, the guest init, loads the NVIDIA modules there at guest boot.

Tested on bare-metal EPYC (Turin) + H200 with that binding (plus `amd_iommu=on iommu=pt`): the device sits under `/sys/bus/pci/drivers/vfio-pci` with its IOMMUFD cdev — which is what the `nvidia-vfio-cdi` service enumerates into the CDI spec — the only extensions installed were the kata CoCo ones (no NVIDIA kmod or toolkit extensions), and `nvidia-smi` runs inside the SEV-SNP guest.

calintkd:
For reference, the host-untouched GPU is also NVIDIA's own architecture for this workload class, not something this extension invents: in the GPU Operator's `vm-passthrough` mode the node gets the "VFIO Manager - to load vfio-pci and bind it to all GPUs on the node" instead of the datacenter driver ([GPU Operator with KubeVirt](https://docs.nvidia.com/datacenter/cloud-native/gpu-operator/latest/gpu-operator-kubevirt.html)), and NVIDIA's confidential-computing tooling runs GPU attestation from inside the guest — the host side only configures GPU CC mode and launches the VM ([nvtrust](https://github.com/NVIDIA/nvtrust)). The machine-config vfio binding above is the Talos-native equivalent of that vfio-manager step.

frezbo:
use PCIDriverRebindConfig since `.machine.kernel.modules` may not guarantee the vfio is always applied if vfio gets auto loaded by udev as a dependency.

We'll then use the instructions to add in test in talos that this works

calintkd:
Done — the README now leads with `PCIDriverRebindConfig`:

```yaml
machine:
  kernel:
    modules:
      - name: vfio_pci
---
apiVersion: v1alpha1
kind: PCIDriverRebindConfig
name: 0000:c1:00.0 # the GPU's PCI address
targetDriver: vfio-pci
```

The parameter-free `vfio_pci` entry stays only so the driver is registered — the rebind writes `driver_override` and pokes `drivers_probe`, which binds only an already-loaded driver; and unlike the `ids=` module parameter, the rebind binds the device no matter what loaded the module first. (One note for the test: `name` is interpolated verbatim into `/sys/bus/pci/devices/%s`, so it needs the sysfs form with the single-digit function — `0000:c1:00.0`; the doc example's `0000:04:00.00` doesn't resolve in sysfs.)

container-runtime/kata-containers-nvidia-gpu-snp/nvidia-vfio-cdi/main.go
```diff
@@ -0,0 +1,125 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
```
frezbo: is there a pod or something from kata side that would write this?

calintkd: Kata doesn't ship a host-side writer — the runtime consumes the host CDI spec (the only CDI generation in kata's flow is inside the guest, where NVRC runs `nvidia-ctk cdi generate` for the guest spec). On the NVIDIA side, yes: the sandbox-device-plugin the GPU Operator deploys for Kata vm-passthrough nodes writes `/var/run/cdi/nvidia.yaml` with `kind: nvidia.com/pgpu` — that's the flow kata's GPU-passthrough doc describes.

This service is the boot-time equivalent for nodes that don't run the GPU Operator: the spec exists as soon as the node is up, and the extension stays self-contained — the same reasoning as the toolkit extension shipping `nvidia-cdi-gen` as an extension service on the host-driver path. On Talos the operator's sandbox flow is largely covered anyway (`PCIDriverRebindConfig` does the vfio binding its vfio-manager exists for), and a node that does run the operator's plugin wouldn't need (or want) both writers. The spec shape is the same either way — `nvidia.com/pgpu`, IOMMUFD cdev device names, cdev-only so one name maps to one GPU — and it's what the SNP+H200 validation ran.

frezbo: we're going to drop nvidia-cdi-gen in favor of upstream NRI support, hence asking if we can just re-use operator, trying to be upstream as much as possible

calintkd: Makes sense — dropped the `nvidia-vfio-cdi` service; the extension is now payload-only (the Go helper, its service spec, and the MPL-2.0 sbom entry are gone). The operator's kata sandbox device plugin (v0.0.3+) generates the same cdev-only `nvidia.com/pgpu` spec on IOMMUFD hosts, so it fully covers this — the README now documents that setup (`sandboxWorkloads.enabled=true`; `vfioManager.enabled=false`, since `PCIDriverRebindConfig` does the binding; `kataManager` off, since the extension ships the runtime).

container-runtime/kata-containers-nvidia-gpu-snp/pkg.yaml
```diff
@@ -0,0 +1,142 @@
...
+      cpes:
+        - cpe:2.3:a:katacontainers:kata_containers:{{ .KATA_CONTAINERS_VERSION }}:*:*:*:*:*:*:*
+      # Apache-2.0: kata-agent (in the guest image), NVRC. GPL-2.0: QEMU
+      # and its datadir blobs, guest kernel, guest image userspace.
+      # LGPL-2.1: libseccomp statically linked into the kata-agent (kata's
+      # own release notice). CUSTOM: NVIDIA proprietary userspace + GSP
+      # firmware in the guest image (the string the nonfree-kmod-nvidia
+      # sbom in siderolabs/pkgs uses for the driver).
+      licenses:
+        - Apache-2.0
```
frezbo, line 134: @dsseng does this feel right?

container-runtime/kata-containers-nvidia-gpu-snp/pkg.yaml
```diff
@@ -0,0 +1,142 @@
...
+        cp ./opt/kata/share/kata-containers/vmlinuz-nvidia-gpu.container /rootfs/usr/local/share/kata-containers/vmlinuz-nvidia-gpu.container
+        cp ./opt/kata/share/kata-containers/kata-containers-nvidia-gpu-confidential.img /rootfs/usr/local/share/kata-containers/kata-containers-nvidia-gpu-confidential.img
+      - |
+        mkdir -p /rootfs/etc/cri/conf.d
+        cp /pkg/12-kata-containers-nvidia-gpu-snp.part /rootfs/etc/cri/conf.d/12-kata-containers-nvidia-gpu-snp.part
+      # The guest image embeds NVIDIA's proprietary userspace driver; NVIDIA's
+      # driver license permits redistribution only if the agreement text
+      # accompanies the binaries, and the kata guest image ships none. Vendored
+      # verbatim from the driver's own LICENSE (as also shipped by the
+      # nvidia-container-toolkit extension under NVIDIA_GLX-1.0/LICENSE).
```
frezbo: this comment seems wrong? i don't see the container-toolkit extension being used?

calintkd: Right — it isn't used here; the comment was citing it as precedent (the toolkit ships the driver license the same way for its host driver stack). Reworded to say that explicitly.

container-runtime/kata-containers-nvidia-gpu-snp/pkg.yaml
dsseng: Wait, these are CUSTOM as they should be: https://github.com/siderolabs/pkgs/blob/main/nonfree/kmod-nvidia/lts/pkg.yaml

And in extensions we specify nothing, since they aren't the packages contributing any executables

dsseng: Overall the combination sounds plausible (and the guest image should be covered, since it's redistributed here), but these licenses cannot be combined with neither OR (each component is under its own), nor AND (they contradict each other)

But I guess as long as all of them will land in the final SBOM the user will be acknowledged what's inside and can review it deeper

calintkd:
Correction to the description and the in-tree docs: Talos v1.13.7 backported the kernel options this extension needs (`CONFIG_IOMMUFD` + `CONFIG_VFIO_DEVICE_CDEV=y`, siderolabs/pkgs#1608 / siderolabs/talos#13765), so the compatibility floor is >= v1.13.7, not v1.14 as the README, the `manifest.yaml.tmpl` comment, and the description said. Amended the commit to correct those two files and edited the description to match; no functional change — the encoded `>= v1.12.0` placeholder and its mechanics are untouched.

frezbo:
okay, is this tested on a real system, it seems it's hard for us to have a test harness in cloud for this setup

calintkd:
Yes — bare-metal AMD EPYC (Turin) with an H200. The stack built from this branch booted an SEV-SNP guest with the GPU passed through on the IOMMUFD cdev, `nvidia-smi` green inside the encrypted guest with GPU CC mode on, and both SNP launch measurements matched computed references. That run was on a `CONFIG_IOMMUFD=y` kernel; the `=m` + allowlisted `iommufd.ko` shape that v1.13.7 and v1.14 actually ship I validated separately on the same box — siderolabs/pkgs#1608.

One thing to be precise about: those runs predate the #1155 restructure, so the SNP base was still self-contained rather than layered on `kata-containers`. The payload is the same tarball objects and the generated config is byte-identical, and every path the handler resolves — shim and virtiofsd from `kata-containers`, `AMDSEV.fd` from `kata-containers-snp` — is present in those extensions, but the three-extension composition itself hasn't been booted as a unit. The CDI spec in those runs also came from the service this PR used to ship rather than the operator's plugin, as the description notes; same cdev-only `nvidia.com/pgpu` shape, and no CDI writer ships here either way.

container-runtime/kata-containers-nvidia-gpu-snp/manifest.yaml.tmpl
```diff
@@ -0,0 +1,19 @@
+version: v1alpha1
+metadata:
+  name: kata-containers-nvidia-gpu-snp
+  version: "{{ .VERSION }}"
+  author: Calin Stefan
+  description: |
+    [{{ .TIER }}] This system extension provides the kata-qemu-nvidia-gpu-snp containerd runtime handler: Kata Containers AMD SEV-SNP confidential VMs with NVIDIA GPU VFIO passthrough (the GPU driver runs inside the encrypted guest, never on the host). Requires the kata-containers and kata-containers-snp extensions from the same release on the same node.
+  compatibility:
+    talos:
+      # PLACEHOLDER floor (inherited from kata-containers-snp): the real
```
frezbo: can we drop placeholders

calintkd:
While validating this on hardware I found a defect in my own README: the GPU Operator recipe here could not have worked as written.

It set `sandboxWorkloads.enabled=true` but not `mode=kata`. The chart default is `kubevirt`, which deploys `sandboxDevicePlugin` — the KubeVirt plugin — rather than the `kataSandboxDevicePlugin` that writes the CDI spec; neither `mode` nor that key exists before gpu-operator v26.3.0, which the text also didn't say. On the default the node never advertises `nvidia.com/pgpu`, so the pod in the Testing section cannot schedule. And the check I documented was the wrong one either way: the Kata plugin advertises `nvidia.com/pgpu` whether or not it wrote a spec, so a node with no spec passes it and fails later at sandbox creation with `unresolvable CDI devices`. It now checks for the spec file, whose name was also wrong: `nvidia.com-pgpu.yaml`, not `nvidia.yaml`. RuntimeClass overhead goes 2Gi → 10Gi to match upstream kata-deploy's budget for this handler, and the build now asserts `cold_plug_vfio`, `pod_resource_api_sock` and `default_memory` from the generated config.

On the hardware question from before: the three extensions have now run as a unit. Bare-metal EPYC 9355 with an H200 NVL, Talos v1.13.8, all built locally from this branch with bldr. Two runs — the first needed a hand-written CDI spec because the sandbox device plugin in place was too old to emit one; the second, with the operator writing its own, came up clean: guest kernel `6.18.35-nvidia-gpu`, `SEV: SNP running at VMPL0`, `nvidia-smi` reporting the H200 NVL, and `nvidia-smi conf-compute -f` returning `CC status: ON`.

container-runtime/kata-containers-nvidia-gpu-snp/manifest.yaml.tmpl
calintkd: Dropped. Manifest is now a bare `version: ">= v1.13.0"` — the highest the pinned validator accepts. Also removed the placeholder wording from the README and PR description. The real v1.13.7 kernel requirement stays in the README.

frezbo:
> While validating this on hardware I found a defect in my own README: the GPU Operator recipe here could not have worked as written.
> 
> It set `sandboxWorkloads.enabled=true` but not `mode=kata`. The chart default is `kubevirt`, which deploys `sandboxDevicePlugin` — the KubeVirt plugin — rather than the `kataSandboxDevicePlugin` that writes the CDI spec; neither `mode` nor that key exists before gpu-operator v26.3.0, which the text also didn't say. On the default the node never advertises `nvidia.com/pgpu`, so the pod in the Testing section cannot schedule. And the check I documented was the wrong one either way: the Kata plugin advertises `nvidia.com/pgpu` whether or not it wrote a spec, so a node with no spec passes it and fails later at sandbox creation with `unresolvable CDI devices`. It now checks for the spec file, whose name was also wrong: `nvidia.com-pgpu.yaml`, not `nvidia.yaml`. RuntimeClass overhead goes 2Gi → 10Gi to match upstream kata-deploy's budget for this handler, and the build now asserts `cold_plug_vfio`, `pod_resource_api_sock` and `default_memory` from the generated config.
> 
> On the hardware question from before: the three extensions have now run as a unit. Bare-metal EPYC 9355 with an H200 NVL, Talos v1.13.8, all built locally from this branch with bldr. Two runs — the first needed a hand-written CDI spec because the sandbox device plugin in place was too old to emit one; the second, with the operator writing its own, came up clean: guest kernel `6.18.35-nvidia-gpu`, `SEV: SNP running at VMPL0`, `nvidia-smi` reporting the H200 NVL, and `nvidia-smi conf-compute -f` returning `CC status: ON`.

so is this good to go?

calintkd:
Good to go from my side — pushed a few README fixes after you asked, docs only, no build changes.

frezbo (approved):

Issue #431, opened 2024-07-22
Support Sysbox


stevefan1999-personal:
https://github.com/nestybox/sysbox/blob/master/docs/user-guide/install-k8s.md#installation-of-sysbox

Another runtime class other than Kata and gVisor :) But this time more close to vanilla runC experience and focus more on light isolation of containers like reporting actual allocated CPU and memory information, so that it won't always hit cgroups OOM manager or something like that. You can also run systemd on Sysbox though I don't think that would work in Talos, since systemd isn't used in Talos.

stevefan1999-personal:
no-stale

stevefan1999-personal:
We need to enable [CRI-O](https://kubernetes.io/docs/setup/production-environment/container-runtimes/#cri-o) support first

yyewolf:
sysbox now supports containerd, this could be done without cri-o support

PR #1171, opened 2026-07-28
feat: add sysbox extension support

yyewolf:
This is an attempt to resolve the following issue: #431 

Here's my use case for sysbox: 

I self-host CI at home for various projects (private and public) and have a need to be able to run exotic workload during those jobs (notably kind in kubernetes)

I tried to do using Kata or gvisor but the restrictions (especially on networking) are making the test environment flakier and also unrealistic

---

This is only a draft and a request to make this real, however I am unsure about the static pod approach to run the sysbox components.

I successfully run this exact setup on my cluster without any issues and this is the kind of stuff I'm trying to run: https://woodpecker.yewolf.fr/repos/5/pipeline/316/5 (this run is on talos with the extension)


MAINTAINERS.md
```diff
@@ -62,6 +62,7 @@ If the field is marked as `Needs Maintainer`, it means that the package is curre
 | revpi-firmware                            | Martin Schuessler  | [c0ffee](https://github.com/c0ffee)                                                      |
 | spin                                      | Sven Pfennig       | [0xE282B0](https://github.com/0xE282B0)                                                  |
 | stargz-snapshotter                        | Sidero Labs        | NA                                                                                       |
+| sysbox                                    | Needs Maintainer   | NA                                                                                       |
```
shanduur: Contrib tier extensions require community maintainers - if you are submitting it here, you should make yourselves a maintainer. We can try to keep the extension up-to-date, but in case of any big issue, broken build, etc., you would need to step up or it will be disabled. Without maintainer, we won't merge it.

yyewolf: I wasn't sure, I'll put myself as a maintainer then.

container-runtime/sysbox/pkg.yaml
```diff
@@ -0,0 +1,122 @@
...
+      # the contents of the sysbox data stores (e.g. /var/lib/docker of a system
+      # container) in and out of the container
+      - url: https://download.samba.org/pub/rsync/src/rsync-{{ .RSYNC_VERSION }}.tar.gz
+        destination: rsync.tar.gz
+        sha256: {{ .RSYNC_SHA256 }}
+        sha512: {{ .RSYNC_SHA512 }}
+      # sysbox-fs mounts its per-container FUSE filesystem by executing
+      # fusermount3, which it looks up in PATH
+      - url: https://github.com/libfuse/libfuse/releases/download/fuse-{{ .FUSE3_SYSBOX_VERSION }}/fuse-{{ .FUSE3_SYSBOX_VERSION }}.tar.gz
+        destination: fuse.tar.gz
```
smira, line 60: fuse should be already available as a separate extension, is there a reason to dupliate?

container-runtime/sysbox/pkg.yaml
```diff
@@ -0,0 +1,122 @@
...
+        sha256: {{ .FUSE3_SYSBOX_SHA256 }}
+        sha512: {{ .FUSE3_SYSBOX_SHA512 }}
+    prepare:
+      - |
+        mkdir -p sysbox-ce rsync fuse
+
+        # the sysbox packages ship statically linked binaries (they are built
+        # with `make sysbox-static`), so they can be used as-is on Talos
+        ar p sysbox-ce.deb data.tar.xz | tar -xJ -C sysbox-ce \
```
smira: is there any reason we can't build sysbox from source? any problem with the toolchain?

container-runtime/sysbox/pkg.yaml
yyewolf: It needs it in a particular location and I couldn't quite figure out how to do it effectively

Should I document the dependency and use `ln` or a similar concept ?

yyewolf: Oh and also, sysbox uses fuse3 in its sysbox-fs process which runs in the static pod container which would need binding with Talos' libc since the fuse3 extension is not statically built

container-runtime/sysbox/pkg.yaml
yyewolf: No I kinda went the easy way since their pipeline used to be more annoying, but it wasn't that bad, it's better now though

container-runtime/sysbox/pkg.yaml
smira: I'm lost on this, if it runs it on the host, host has the libs; if it runs within the pod, why does it need on the host (in the extension). Neither makes sense to me.

yyewolf: Because no image is built for the pod, the pod is just there to be in the same PID namespace as the kubelet which I didn't figure out how to do with a normal Talos service

smira: I'm not sure what is a pod there in this description. It either runs within the host, or not. Also we have `gvisor` already which works without any problem.

yyewolf: It's multiple things:

`gvisor` doesn't have long-running daemons though in its model it adds additional processes to the sandboxes of containers and the CRI talks to them directly since it originated them.

`sysbox-runc` on the other hand relies on `sysbox-mgr` and `sysbox-fs` to be preexisting processes and communicates with them. When `sysbox-runc` creates a sandbox, it gets hosts PIDs and communicates those to `sysbox-mgr` and `sysbox-fs`.

Talos runs extension services in a private PID namespace, but `sysbox-mgr` AND `sysbox-fs` are handed host PIDs by `sysbox-runc` and they do calls like `pidfd_open(2)` and also rely on seccomp notifications on those PIDs.

Since those PIDs are hosts PIDs, `sysbox-mgr` AND `sysbox-fs` need to be on the host PID namespace, which is not possible through Talos' services. That's where the decision of having static pods comes from.

The pods have no images because the build process doesn't have a way of building images for those, so that's why it's relying on the rootfs to have the necessary binaries to run properly.

smira: Talos 1.14 does CRI sandboxing, if you mean that, but that is optional.

Static pods is not a great ideas, as it requires e.g. kubelet to be running, so in theory kubelet can't use a new container runtime. 

All in all, idk is it too painful and hacky? Does it have enough value to fit in between e.g. gvisor and kata? Can't you just use kata instead? Yes, it has bigger overhead, but less hacks, and truly good isolation?

yyewolf: No I do not mean CRI sandboxing, but yeah CRI sandboxing will pose another challenge to run sysbox.

Also Kata's problem is not overhead, Kata cannot run every workload the way Sysbox can.

For example Kata's PID 1 inside the microVM is `kata-manager` (or something like that) meaning you cannot run `systemd` inside a container, this is something you can easily do using Sysbox because of fs emulation and partial kernel emulation. Same for `dockerd`, overlayfs, iptables and cgroups issues are all handled by those features.

Supposedly, [vnode](https://www.vnode.com/) would also solve the same problem, but it probably will be following the same licensing as vCluster

PR #1183, opened 2026-08-04
feat: add cachefilesd extension

ChristianJacobsen:
Adds cachefilesd: NFS mounts with the `fsc` option cache reads on local disk.

Depends on: siderolabs/pkgs#1636


storage/cachefilesd/cachefilesd.yaml
```diff
@@ -0,0 +1,72 @@
...
+        - ro
+    # libc
+    - source: /usr/lib
+      destination: /usr/lib
+      type: bind
+      options:
+        - bind
+        - ro
+    # cachefilesd binary
+    - source: /usr/local/sbin
```
smira: do we need it to be on the host? or can it be self-contained contianer rootfs?

storage/cachefilesd/cachefilesd.yaml
```diff
@@ -0,0 +1,72 @@
...
+        - ro
+    # config file
+    - source: /usr/local/etc/cachefilesd.conf
+      destination: /usr/local/etc/cachefilesd.conf
+      type: bind
+      options:
+        - bind
+        - ro
+    # cache root, rw: the cachefiles kernel module sets xattrs on it
+    - source: /var/cache/fscache
```
smira: I understand that Talos at the moment doesn't support a nice way to handle this, but it almost feels like this is too restrictive, e.g. this doesn't allow a user volume to be used for the cache. 

I might need to think a bit more about this specific usecase

storage/cachefilesd/cachefilesd.yaml
```diff
@@ -0,0 +1,72 @@
...
+        - ro
+    # cachefilesd binary
+    - source: /usr/local/sbin
+      destination: /usr/local/sbin
+      type: bind
+      options:
+        - bind
+        - ro
+    # config file
+    - source: /usr/local/etc/cachefilesd.conf
```
smira: Talos now supports custom `EtcFileConfig` to allow users to override configuration, I wonder if the extension shouldn't ship any, and instead bind-mount the file from `/etc` which is user-managed...

smira (commented):
please add yourself to MAINTAINERS for this extension

storage/cachefilesd/cachefilesd.yaml
ChristianJacobsen: Yeah, that is probably the better solution. I mirrored what was done in `lldpd` so everything is in a self-contained rootfs now.

storage/cachefilesd/cachefilesd.yaml
ChristianJacobsen: Agreed it's restrictive. `/var` being ephemeral works for my use case, but yes - it would be strictly better with a user volume. I'm not hitched to any solution, so I'll go with whatever suggestion you think is best. You know this stack better than I.

storage/cachefilesd/cachefilesd.yaml
ChristianJacobsen: Looked at this and ran into a wrinkle: machined `os.MkdirAlls` any missing bind-mount source, so a bind-mounted `/etc/cachefilesd.conf` would be created as a directory when the user doesn't supply one, which would break the daemon. So the default config now lives in the container rootfs, and users can override it via `ExtensionServiceConfig`. I'm happy to switch to depends: configuration: true with no defaults if you prefer that instead.

ChristianJacobsen:
I'll squash any fixups before merge, but it's easier to track the fixes this way instead of force-pushing.

storage/cachefilesd/cachefilesd.yaml
frezbo: you need to add a file dependency, so when etcfileconfig creates only the service tries to start

ChristianJacobsen: Ahh, I see. That makes more sense. I hope I got it right this time. At least it works "as intended" in a Qemu cluster on my Mac 😄

storage/cachefilesd/cachefilesd.yaml
smira: I reworked it to use a UserVolume and updated README

smira (approved):

smira:
> I'll squash any fixups before merge, but it's easier to track the fixes this way instead of force-pushing.

I reworked/simplified things to use user volumes which should be way more flexible.

ChristianJacobsen:
@smira, yeah this looks much better! 😄

Anything more you need from me?

smira:
> @smira, yeah this looks much better! 😄
> 
> Anything more you need from me?

no, thanks, waiting for internal review/timing, but it should land in 1.14.0 anyways

storage/cachefilesd/README.md
```diff
@@ -0,0 +1,61 @@
+# cachefilesd
+
+This extension provides [cachefilesd](https://www.kernel.org/doc/html/latest/filesystems/caching/cachefiles.html) - the userspace daemon for FS-Cache, the Linux kernel filesystem caching facility. It ships both the `cachefiles` kernel module and the `cachefilesd` daemon.
+
+FS-Cache allows network filesystems (NFS, Ceph, etc.) to cache data on local disk, dramatically reducing cold-start latency for read-heavy workloads. The cache is transparent: applications see no change in behavior, but repeated reads from the network filesystem are served from local disk instead.
+
+## Usage
+
+The kernel module is not loaded automatically. Add `cachefiles` to `.machine.kernel.modules` in your machine config:
```
frezbo:
```suggestion
The kernel module is not loaded automatically. Add `cachefiles` to `KernelModuleConfig` in your machine config:
```

frezbo (approved):
🆒

storage/cachefilesd/README.md
smira: fixed, thanks!

PR #1188, opened 2026-08-05
chore: replace gnu mirror

shanduur:
ftpmirror.gnu.org is down so replace it with ftp.gnu.org

### DIFF
--- a/misc/glibc/pkg.yaml
+++ b/misc/glibc/pkg.yaml
@@ -17,7 +17,7 @@ steps:
       LD_LINUX_X86_64: ld-linux-x86-64.so.2
       LD_LINUX_AARCH64: ld-linux-aarch64.so.1
     sources:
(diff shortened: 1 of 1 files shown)


misc/glibc/pkg.yaml
```diff
@@ -17,7 +17,7 @@ steps:
       LD_LINUX_X86_64: ld-linux-x86-64.so.2
       LD_LINUX_AARCH64: ld-linux-aarch64.so.1
     sources:
-      - url: https://ftpmirror.gnu.org/gnu/glibc/glibc-{{ .GLIBC_VERSION }}.tar.xz
+      - url: https://ftp.gnu.org/gnu/glibc/glibc-{{ .GLIBC_VERSION }}.tar.xz
```
frezbo, line 20: pick https://github.com/siderolabs/extensions/pull/1186

PR #1191, opened 2026-08-08
feat: Added wpa-supplicant extension

kurtassad:
### What this does

Adds the `wpa-supplicant` system extension so Talos nodes can join wifi networks. 

Credentials are supplied at runtime via `ExtensionServiceConfig` with the SSID/PSK baked in. The service is hardcoded to `wlan0`, so images are required to use `--extra-kernel-arg net.ifnames=0`.

This extension alone is not enough for WiFi. `wpa_supplicant` talks to the kernel but needs the `cfg80211` submodule plus a WiFi driver if not upstreamed into linux mainline already. Users need a custom kernel that includes them.

I am not proposing to upstream the kernel changes, they are easy enough to configure, but should probably not be upstreamed due to the increased attack surface. But I think that shipping the extension would be nice because it takes out 99% of the work (enabling the submodule is just a few line change and difficult to screw up).

Reference kernel change is here:
https://github.com/kurtassad/pkgs/commit/9e7dd1d99accdfd179e08d03073ccc4bb17c7e8b

## Testing

- I built the extension image
- Baked into Talos with `--extra-kernel-arg net.ifnames=0` onto the kernel refernece above that enables CFG80211.
- Applied KernelModuleConfig and ExtensionServiceConfig in accordance with the readme in this PR
- Confirmed that the setup worked end to end on latest talos. Basically ext-wpa-supplicant runs and associates with `wlan0` and gets the DHCP lease.


smira:
We do not have immediate plans to support WiFi in Talos, as it requires more support in core, and WiFi doesn't seem to the most reliable cluster networking.

PR #1206, opened 2026-08-17
fix: update broken talos.dev URLs to docs.siderolabs.com

pranav767:
Updated boot-assets and system-extensions URLs across multiple README files to point to the correct documentation location.


smira (approved):
thank you!

PR #1224, opened 2026-09-01
fix: avoid virtqemud reconnect deadlock

frezbo:
Talos provides a system D-Bus socket but does not run systemd-machined. virtqemud blocks while probing that unavailable service during QEMU domain reconnect, leaving management RPCs hung after a daemon restart.

Point virtqemud at an absent bus socket so libvirt immediately treats systemd integration as unavailable.


dsseng (approved):
What purposes does virtqemud connect to systemd for? Could it be (like kubelet) listening for some events that might be of importance to us like power events?

PR #1231, opened 2026-09-03
feat: replace wolfi build roots with stagex

frezbo:
Build glibc and NVIDIA userspace from pinned Sidero-controlled inputs instead of resolving packages from Wolfi repositories. Keep the runtime layout intact while introducing a glibc sysroot for CGO and C builds.

Tested with NVIDIA tests manually.


smira (approved):
let's try it out!

PR #1238, opened 2026-09-06
feat: add dvb-pt3 extension

5ym:
## What

A `dvb-pt3` system extension for the [Earthsoft PT3](https://earthsoft.jp/PT3/), an ISDB-T/S PCIe tuner card.

## Why

The PT3 is the common choice for terrestrial and satellite TV recording in Japan, and it is one of the last things keeping such a workload off Talos. The driver is in-tree; nothing out-of-tree is built here.

## How

Modelled directly on `dvb/cx23885`, minus the firmware download: the PT3 needs no firmware blob. The extension ships `earth-pt3.ko` and what it pulls in.

| module | role |
| --- | --- |
| `earth-pt3` | the card |
| `tc90522` | demodulator |
| `qm1d1c0042` | ISDB-S tuner |
| `mxl301rf` | ISDB-T tuner |
| `dvb-core`, `mc` | DVB core and media controller |

Module paths and the dependency chain were taken from a running host with the card installed (`modinfo -n`, `lsmod`), not guessed. Modules are blacklisted by default like `dvb-cx23885`, so they load only when the machine config asks for `earth-pt3`.

x86_64 only, matching the kernel side: the card is PCIe and there is no practical arm64 use.

## Depends on

siderolabs/pkgs#1682 enables `CONFIG_DVB_PT3=m`. Until that lands and a Talos release picks it up, the module does not exist in the kernel image and this build cannot succeed. The `compatibility.talos.version` in the manifest is set to `>= v1.15.0` as a guess at the release that will carry it, and should be corrected to whatever it actually turns out to be.

## Notes

- The README row was added by hand in the generated block; regenerate if that matters.
- I could not run the extension build locally, so this relies on CI.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01JgGYf5qbjDXYhtzcpLsGvv


smira:
@5ym I did some changes, and added you to `MAINTAINERS.md` - if the entry looks wrong, please feel free to submit a follow-up PR.

smira (approved):

PR #1246, opened 2026-09-10
feat: add urunc container runtime extension

kastakhov:
## What is urunc?

[urunc](https://urunc.io/) is an OCI-compatible low-level container runtime for running unikernels in cloud-native environments. It integrates with containerd and Kubernetes through a containerd shim and a dedicated RuntimeClass.

Upstream urunc supports OCI-packaged Unikraft, Rumprun, MirageOS, Mewz, Hermit, and Linux workloads. This extension bundles the QEMU, Firecracker, Cloud Hypervisor, and Solo5 monitors required by those workloads.

## What does this change?

This PR adds urunc v0.8.0 as a Talos system extension.

It:
- Installs urunc, containerd-shim-urunc-v2, virtiofsd, and the supported VM/Sandbox monitor binaries.
- Registers io.containerd.urunc.v2 as the urunc containerd runtime handler.
- Adds documentation for creating a Kubernetes RuntimeClass and configuring urunc on Talos.
- Supports both amd64 and arm64.
- Adds a downstream urunc patch that passes the nameserver from the container’s OCI-mounted /etc/resolv.conf to Unikraft guests, enabling cluster DNS instead of always using the 8.8.8.8 fallback.
- Fixes the duplicated PUSH=true argument in the extension build documentation.

## Testing

 - Tested on Talos v1.13 deployed into Proxmox VE with unikraft unikernel.


container-runtime/urunc/patches/0001-use-oci-resolver-for-unikraft.patch
```diff
@@ -0,0 +1,325 @@
+From bd994aef47959bec3bc3290359d373cba24e548a Mon Sep 17 00:00:00 2001
+From: Ali Mohamed <amx746@gmail.com>
+Date: Thu, 3 Sep 2026 10:11:20 +0300
+Subject: [PATCH] fix(unikraft): set the guest DNS server from the container's
+ resolv.conf
+
+Unikraft guests used a hardcoded DNS server, so name resolution inside
+the unikernel ignored the DNS configuration of the container. Parse the
+nameserver entry from the container's resolv.conf and pass it to the
```
smira, line 9: can we wait for the next release which includes the patch?

container-runtime/urunc/README.md
```diff
@@ -0,0 +1,54 @@
...
+```
+
+## Configuration
+
+Create the configuration under `/var` with a Talos machine configuration patch, then set `URUNC_CONFIG_FILE` so both `urunc` and its containerd shim use it:
+
+```yaml
+machine:
+  files:
+    - path: /var/etc/urunc/config.toml
```
smira, line 28: `machine.files` has been deprecated, please move to `EtcConfig`

container-runtime/vars.yaml
```diff
@@ -56,3 +56,13 @@ YOUKI_ARM64_SHA256: b96c05c2c82f1d20a74b611188fa120894c50a6128f73856bb371604ecb6
 YOUKI_ARM64_SHA512: 99525a9ea4c071d9a8c7cc422bbf1a97140552d178a78cfa1d7e57ef27d2b518c9f6d30a20ce4ccd0b3914b97532e106b6232bd5a83fcb468c735885318e63ff
 YOUKI_AMD64_SHA256: a211a16b6b1a985f8ca61bba9a30b9ae40352d04dd7eca04cdecfe7b98b47038
 YOUKI_AMD64_SHA512: 48d22651ca75be19793adac8cc62f8128fa634d35c2e2bb73c243f51bb634406e1b5399f6032d1fce297d44c2c7fb54b3bba6df66ecd1c8a031b0cb2b55d0fbf
+# renovate: datasource=github-releases extractVersion=^v(?<version>.*)$ depName=urunc-dev/urunc
+URUNC_UPSTREAM_VERSION: 0.8.0
+URUNC_VERSION: 0.8.0-talos.1
+URUNC_SOURCE_SHA256: 6773f666f2ddfa5ec4b52b35a685d76063ad61e28234a78355a354240858685f
+URUNC_SOURCE_SHA512: ac7b46047bb48d8799b347f2da654b69b3f091096cb6a0bcf4f3190fc68b37d062ab445dc71ae52000ca04d2dfe618ddfbd0df37e3a4d020a342becaa0ce5c0c
+URUNC_MONITORS_VERSION: FC-v1.7.0_CLH-v50.0_S5-v0.12.1_VFS_-v1.13.0_QM-v10.1.1-9a44e
```
smira, line 64: this looks like a mess of a version - it's unclear how to support it in the future

container-runtime/urunc/patches/0001-use-oci-resolver-for-unikraft.patch
qdrddr: When is the next release planned @smira?

smira: urunc is not my project

qdrddr: Ah, you meant the next release of Urunc?

kastakhov: > can we wait for the next release which includes the patch?

I don't think so, the v0.8.0 release was only a few weeks ago, and the next one per roadmap is pretty far away.

container-runtime/urunc/README.md
kastakhov: Sure, moved example to `EtcFileConfig`.

container-runtime/vars.yaml
kastakhov: I agree that this looks kinda mess, but in fact this is only pre-build binaries, so we can either build them ourselves or just download the latest one from repo (like I did).
Kata has something similar, except they build it as part of their release cycle and has separeted tarball with static files.

container-runtime/urunc/patches/0001-use-oci-resolver-for-unikraft.patch
kastakhov: Actually, this patch is not a blocker, but it cover the most usual case when you use pre build Unikraft kernel for image. And for my standpoint Unikraft is most usable option from supported unikernels, as other have some limitation with the application types (at least in our usecase). But if rebuild Unikraft kernel, this patch is no necessary. So I guess, we can either leave a note with required steps to use urunc + Unikraft or leave the patch.

container-runtime/vars.yaml
smira: My question is maintenance (which will be on us), and I don't see how to maintain this version - when to upgrade it and how.

qdrddr: @smira, how would you prefer we structure or publish these versions so they’re easier to support? What direction makes the most sense to you to move this forward?

container-runtime/urunc/patches/0001-use-oci-resolver-for-unikraft.patch
qdrddr: @smira, since the next Urunc release is far out, would you be open to merging this patch into our extension now so we’re not blocked? Would you like us to prepare a clean PR that applies the resolver fix and keeps it isolated or what are our next steps to move from here?

container-runtime/vars.yaml
smira: I don't know the details of urunc, but is there something that I can tell Renovate to match on?

container-runtime/urunc/patches/0001-use-oci-resolver-for-unikraft.patch
smira: how far out is urunc release? this extension will not ship before Talos 1.15 (December).

Also it ships so much extra stuff that I need to take another look.

kastakhov: Hi @smira, I discussed this with urunc maintainer. Here is the answer:

> We can have a patch release at the first week of October. I will plan some bug fixes this week to push them in v0.8.1. Otherwise, the next minor release is planned for late November early December.

container-runtime/vars.yaml
kastakhov: I don't think so, at least with existing flow, but as far as I know this currently is being discussed how to change monitors build processes. Not sure about specific decisions yet.

PR #1255, opened 2026-09-15
feat(nfsrahead): read config from /etc/nfs.conf

buroa:
nfsrahead was built with --with-nfsconfig pointing at /usr/local/etc/nfsrahead/nfs.conf, a read-only path inside the extension image, so the readahead values could not be changed per cluster.

Point it at /etc/nfs.conf instead, which is also upstream's default, so the settings can be supplied through an EtcFileConfig machine config document. This additionally enables /etc/nfs.conf.d/*.conf drop-ins, since nfs-utils only scans the drop-in directory when the configured path starts with /etc/.

Extensions cannot ship files under /etc, as extensions-validator rejects any path outside AllowedPaths, so the bundled defaults are dropped rather than relocated. EtcFileConfig landed in Talos v1.14.0 and machine.files writes relative to /var, so the compatibility floor moves to v1.14.0.

BREAKING CHANGE: the nfs=15000 and nfs4=16000 defaults are no longer shipped with the extension. Without an EtcFileConfig providing /etc/nfs.conf, nfsrahead falls back to a readahead of 128. To keep the previous behavior:

```
apiVersion: v1alpha1
kind: EtcFileConfig
name: nfs.conf
mode: 0o644
contents: |
  [nfsrahead]
  nfs=15000
  nfs4=16000
  default=128
```


smira:
I think in general this extension doesn't have README.md, and this would be a great place to document new behavior and provide an example for new `EtcFileConfig`, lgtm otherwise

buroa:
@smira Done, thanks!

smira (approved):

PR #1257, opened 2026-09-20
feat: update DRBD to 9.3.4

bernardgut:
Via https://github.com/siderolabs/pkgs/pull/1693


smira:
thank you, but we need to do it a bit differently.

